PR #2, opened 2019-07-10
WIP - A suggested structure

ashleymichal:
**NOT INTENDED FOR MERGE**, this is intended to serve as a conversation starter.

This is a v. opinionated first pass at a proposal for extending/guiding contributions to this tool. From what I've discerned from a not-too-deep analysis of the codebase, it appears what we have is actually a framework/library for interacting with v4 Cloudflare APIs, and a series of API-specific client modules that use the framework.

I think that treating these things separately can help to establish the contribution boundaries for potential users; exposing Endpoint, APIResponse, etc as parts of an "external" crate, grants the option for folks to either contribute a client module to this repo, or to develop their own client module alongside their APIs and export their own crate, using this framework as dependency.

This PR includes a `client` directory with a `README` which acts as a draft of a usage / contribution guide for developing a client using the framework. It also includes an example of a client module, converted from the existing `zone` module as an example. Providing this structure and README makes the usage clearer to a newcomer. An additional improvement would include sample unit tests for client endpoints and/or data structures.

### DIFF
--- a/src/clients/README.md
+++ b/src/clients/README.md
@@ -1,0 +1,41 @@
+# Building New Client Modules
+
+## TODO:
+
+### add template directory to indicate usage:
+
+* base directory ("/sample") should be named after the API in question.
+* mod.rs should instantiate the internal mods, use them to make them public:
+``` rust
+mod do_something;
+
+pub use do_something::DoSomething // where DoSomething is the Endpoint implementation.
+```
+* mod.rs should define common data structures for the module (i.e. Zone and its children). generally this includes implementation of the APIResult trait for top-level objects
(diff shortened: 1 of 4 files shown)


adamchalmers:
I really like the idea of separating the fundamentals/infra/core into one module, and the actual endpoint implementations (DNS, Workers, WorkersKV, etc) into another module. Agree that it'd make contributing much easier.

adamchalmers:
We agreed to close this in favour of [my PR here](https://github.com/cloudflare/cloudflare-rs/pull/24) which has the same goal.

Issue #6, opened 2019-07-25
Update CreateNamespace endpoint to take params


ashleymichal:
The CreateNamespace endpoint should take a json object containing a `title` key as its POST body.

PR #7, opened 2019-07-25
Alewis/workerskv module

### LINKED ISSUE #5: Restructure WorkersKV module
We'll be testing out use of this api client in https://github.com/cloudflare/wrangler, specifically for interacting with the Workers KV API. This ticket covers a small refactor around the structure of the `workerskv` module.

ashleymichal:
Closes #5 and #6.


ashleymichal:
Looks like the master branch is behind the BB repo by 2 commits.

src/workerskv/mod.rs
```diff
@@ -0,0 +1,38 @@
+use crate::response::APIResult;
+use chrono::offset::Utc;
+use chrono::DateTime;
+
+mod create_namespace;
+mod list_namespace_keys;
+mod list_namespaces;
+mod remove_namespace;
+mod rename_namespace;
```
nickvollmar: Discussed offline: these things aren't `pub` but should be, so callers can construct the relevant Params objects, etc.

src/workerskv/mod.rs
```diff
@@ -2,17 +2,11 @@ use crate::response::APIResult;
...
-pub use create_namespace::CreateNamespace;
-pub use list_namespace_keys::ListNamespaceKeys;
-pub use list_namespaces::ListNamespaces;
-pub use remove_namespace::RemoveNamespace;
-pub use rename_namespace::RenameNamespace;
+pub mod create_namespace;
+pub mod list_namespace_keys;
+pub mod list_namespaces;
+pub mod remove_namespace;
+pub mod rename_namespace;
```
adamchalmers: I'd also be okay with `pub use create_namespace::*` if you want to just move these all into one big WorkersKV namespace.

ashleymichal: was going to asked what y'all preferred, is that the preference?

adamchalmers: Actually now that I've thought about it more, I like this approach because as the number of endpoints implemented gets bigger, this will be more manageable. But given that I don't have a strong opinion either way, and you're the one using the code, I'll defer to whatever you prefer.

PR #9, opened 2019-08-01
purging cache

tigranbs:
I had this strange case for TreeScale (https://github.com/treescale) where I needed to purge cached assets per zone based on my user actions.

Basically added another struct and parameter type for Zone, probably need to add better example implementation

### DIFF
--- a/example/main.rs
+++ b/example/main.rs
@@ -37,6 +37,21 @@ fn print_response<T: APIResult>(response: APIResponse<T>) {
 
 fn zone<APIClientType: APIClient>(arg_matches: &ArgMatches, api_client: &APIClientType) {
     let zone_identifier = arg_matches.value_of("zone_identifier").unwrap();
+    let purge = arg_matches.value_of("purge").unwrap();
+    if purge != "false" {
+        let response = api_client.request(&zone::ZoneCachePurge {
+            identifier: zone_identifier,
+            params: zone::PurgeZoneCacheParams {
+                tags: vec![String::from("")],
+                hosts: vec![String::from("")],
+                files: vec![String::from("")],
+            }
+        });
+
+        print_response(response);
+        return;
+    }
+
     let response = api_client.request(&zone::ZoneDetails {
         identifier: zone_identifier,
     });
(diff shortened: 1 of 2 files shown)


src/zone.rs
```diff
@@ -35,6 +35,33 @@ impl<'a> Endpoint<Zone> for ZoneDetails<'a> {
     }
 }
 
+
+/// Zone Details
+/// https://api.cloudflare.com/#zone-purge-files-by-cache-tags-or-host
+pub struct ZoneCachePurge<'a> {
```
sssilver: To be consistent with documentation and the rest of the codebase, it'd probably be better for this endpoint to be called `ZonePurgeFiles`.

src/zone.rs
```diff
@@ -35,6 +35,33 @@ impl<'a> Endpoint<Zone> for ZoneDetails<'a> {
     }
 }
 
+
+/// Zone Details
```
sssilver: We try to keep these doc comments consistent with the official API documentation endpoint titles. This particular one should be: `/// Purge Files by Cache-Tags or Host`

src/zone.rs
```diff
@@ -35,6 +35,33 @@ impl<'a> Endpoint<Zone> for ZoneDetails<'a> {
...
+    fn path(&self) -> String {
+        format!("zones/{}/purge_cache", self.identifier)
+    }
+    fn body(&self) -> Option<PurgeZoneCacheParams> {
+        Some(self.params.clone())
+    }
+}
+
+#[derive(Serialize, Clone, Debug)]
+pub struct PurgeZoneCacheParams {
```
sssilver: The documentation specifies that both `tags` and `hosts` are optional parameters. The client code should probably reflect that by wrapping them in an `Option`.

src/zone.rs
```diff
@@ -35,6 +35,33 @@ impl<'a> Endpoint<Zone> for ZoneDetails<'a> {
...
+    pub tags: Vec<String>,
+    pub hosts: Vec<String>,
+    pub files: Vec<String>,
```
sssilver: The documentation doesn't appear to have a `files` parameter for this particular endpoint. "Purge files by URL" seems to have one, but it'd be against the philosophy of this client to conflate two endpoints into one.

example/main.rs
```diff
@@ -37,6 +37,21 @@ fn print_response<T: APIResult>(response: APIResponse<T>) {
...
 fn zone<APIClientType: APIClient>(arg_matches: &ArgMatches, api_client: &APIClientType) {
     let zone_identifier = arg_matches.value_of("zone_identifier").unwrap();
+    let purge = arg_matches.value_of("purge").unwrap();
+    if purge != "false" {
+        let response = api_client.request(&zone::ZoneCachePurge {
+            identifier: zone_identifier,
+            params: zone::PurgeZoneCacheParams {
+                tags: vec![String::from("")],
+                hosts: vec![String::from("")],
+                files: vec![String::from("")],
```
sssilver: This seems non-functional. What's the reason for doing this?

PR #16, opened 2019-08-07
fix #15: Add query to `ListNamespaceKeys` WorkersKV endpoint

### LINKED ISSUE #15: Add query to `ListNamespaceKeys` WorkersKV endpoint
The workerskv::ListNamespaceKeys endpoint is missing the query method and the params constructor argument. Also missing `prefix` query parameter.

ashleymichal:


src/workerskv/list_namespace_keys.rs
```diff
@@ -7,6 +7,7 @@ use crate::endpoint::{Endpoint, Method};
 pub struct ListNamespaceKeys<'a> {
     pub account_identifier: &'a str,
     pub namespace_identifier: &'a str,
+    pub params: ListNamespaceKeysParams,
```
sssilver: Why not `&ListNamespaceKeysParams`?

ashleymichal: that's a great question. the suggested code i received (and merged) for the CreateNamespace endpoint did not use a reference for params. I can add an issue and PR to fix that in other places.

sssilver: Awesome!

PR #20, opened 2019-08-08
fix #17: Adhere to Rust naming conventions

### LINKED ISSUE #17: Adhere to Rust naming conventions
We currently fully capitalize acronyms in type names. However, https://rust-lang-nursery.github.io/api-guidelines/naming.html specifies that in Rust the preference for acronyms is:

> In UpperCamelCase, acronyms and contractions of compound words count as one word: use Uuid rather than UUID, Usize rather than USize or Stdin rather than StdIn. In snake_case, acronyms and contractions are lower-cased: is_xid_start.

sssilver:


src/workerskv/mod.rs
```diff
@@ -19,14 +19,14 @@ pub struct WorkersKVNamespace {
     pub title: String,
 }
 
-impl APIResult for WorkersKVNamespace {}
-impl APIResult for Vec<WorkersKVNamespace> {}
+impl ApiResult for WorkersKVNamespace {}
+impl ApiResult for Vec<WorkersKVNamespace> {}
```
ashleymichal: Probably also need to do this for `WorkersKvNamespace`.

sssilver: Yes! Good catch

sssilver: @ashleymichal I can't find any occurrences of that in the current master which I've branched off of.

ashleymichal: i mean `WorkersKVNamespace -> WorkersKvNamespace`, which is present here.

sssilver: Done!

PR #24, opened 2019-08-08
fix #23: separate modules for endpoints and framework

### LINKED ISSUE #23: Separate modules for the API framework and endpoint implementations
Currently our main `src/` directory contains two kinds of modules: 

- modules for the API framework (e.g. `response` or `apiclient`
- modules for requests/responses from particular Cloudflare APIs (e.g. `dns` or `workerskv`)

We should sort these into two different modules (my suggested names: `framework` and `endpoints`). That way, most people can ignore the framework directory and only focus on adding their own endpoints.

adamchalmers:
I've reorganized the root of the repo. Previously there were two kinds of root modules: modules for particular Cloudflare endpoints (e.g. DNS or WorkersKV) and modules for the fundamental API framework (e.g. response, apiclient, auth).

This PR restructures the root of the crate so there are only _two_ root modules: `framework` and `endpoints`.

Before:
```
src/
    account.rs
    apiclient.rs
    auth.rs
    dns.rs
    lib.rs
    mock.rs
    plan.rs
    response/
    workerskv/
    zone.rs
```

After:

```
src/
    endpoints/
        account.rs
        dns.rs
        plan.rs
        workerskv/
        zone.rs
    framework/
        apiclient.rs
        auth.rs
        mod.rs
        mock.rs
        response/
```

This will make it much more clear for contributors where their code should go. Are you adding a new endpoint? `endpoints` module. Are you changing the underlying API framework? `framework` module.

I've already integrated these changes into a branch of tunnelmon, and it was very straightforward.


sssilver:
@ashleymichal has a PR up that adds even more taxonomy to this.

adamchalmers:
Added a CONTRIBUTING.md!

Issue #26, opened 2019-08-14
HttpApiClient should be able to handle non-JSON request and response bodies


gabbifish:
There are some APIs at Cloudflare that do not follow our common use of json for request and response bodies: Two examples of this:
1. Reading a Workers KV value by key returns a raw string body: https://api.cloudflare.com/#workers-kv-namespace-read-key-value-pair
2. Setting a Workers KV key-value pair requires a raw string body: https://api.cloudflare.com/#workers-kv-namespace-write-key-value-pair

Because cloudflare-rs expects to serialize/deserialize JSON for all CF API endpoints, these Workers KV API endpoints are not compatible with this library.

Given that this library surfaces an HttpApiClient, not an HttpJsonApiClient, should this trait be built more extensibly, so the parsing of request bodies is directly tied to the `"content-type"` header associated with it? Perhaps the default serialization/deserialization will be for json, but if a request/response specifies a specific `"content-type"` header, the body bytes will be parsed as such.

adamchalmers:
I agree. JSON is a sensible default, but if an API specifically tells us to parse something as plain-text or octet-stream, we should skip the JSON deserialization and just return a String.

adamchalmers:
BTW you can send a string body by just defining an `Endpoint` struct where `BodyType = String`.

ashleymichal:
yeah we'll need one where BodyType is a multipart form of a v. specific shape.

ispivey:
@adamchalmers I'm learning myself a lil' Rust and played around with implementing some new APIs, including [Upload a Worker](https://api.cloudflare.com/#worker-script-upload-worker) which requires a string body. It seems like if you define an `Endpoint` struct where `BodyType = String` as you suggested above, the later call to `serde_json::to_string(&body).unwrap()` in `framework::ApiClient::request` will wrap the body in quotes. I made a simple example that illustrates:
1. An endpoint with a `BodyType` that is a struct that contains a string parameter,
2. An endpoint with a `BodyType = String`,
3. An endpoint with that simply prints the string body, without calling `serde_json::to_string` on it.

This results in:
```
1: {"script":"addEventListener('fetch', event => { event.respondWith(fetch(event.request)) })"}
2: "addEventListener('fetch', event => { event.respondWith(fetch(event.request)) })"
3: addEventListener('fetch', event => { event.respondWith(fetch(event.request)) })
```

[...]

PR #27, opened 2019-08-15
[WIP] Initial attempt at making map_api_response accept non-json api responses

gabbifish:
This PR explores a PoC for one part of #26. It focuses on allowing non-json response bodies to be accepted by cloudflare-rs and returned using the same ApiSuccess.result accessor pattern used by preexisting users of cloudflare-rs.

My questionable first approach to this problem: trying to shove the raw text API response into the ApiSuccess.result field, which should ensure that this change is not breaking, but also fails with
```
error[E0308]: mismatched types
  --> /home/gabbi/github/cloudflare-rs/src/response/mod.rs:39:33
   |
39 |                         result: body, 
   |                                 ^^^^ expected type parameter, found struct `std::string::String`
   |
   = note: expected type `ResultType`
              found type `std::string::String`

```
^ I'm not sure how to get around this (or if this approach even makes much sense in the first place). I floated other ideas in the comments in the code changes.

### DIFF
--- a/src/response/apifail.rs
+++ b/src/response/apifail.rs
@@ -46,6 +46,10 @@ impl fmt::Display for ApiError {
 
 pub trait ApiResult: DeserializeOwned + Debug {}
 
+// Ensure that even a raw text response body can be an ApiResult (this takes advantage
+// of the fact that Serde can deserialize strings into... strings).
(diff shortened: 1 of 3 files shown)


kennetheversole:
Hello, 

PR #28, opened 2019-08-15
Add support for Workers KV bulk upload API endpoint

gabbifish:
This PR adds support for the Workers KV bulk upload API endpoint: https://api.cloudflare.com/#workers-kv-namespace-write-multiple-key-value-pairs.

It is used by https://github.com/cloudflare/wrangler/pull/445

### DIFF
--- a/src/endpoints/workerskv/write_bulk.rs
+++ b/src/endpoints/workerskv/write_bulk.rs
@@ -1,0 +1,39 @@
+use crate::framework::endpoint::{Endpoint, Method};
+
+/// Write Key-Value Pairs in Bulk
+/// Writes multiple key-value pairs to Workers KV at once.
+/// A 404 is returned if a write action is for a namespace ID the account doesn't have.
+/// https://api.cloudflare.com/#workers-kv-namespace-write-multiple-key-value-pairs
+pub struct WriteBulk<'a> {
+    pub account_identifier: &'a str,
+    pub namespace_identifier: &'a str,
+    pub bulk_key_value_pairs: Vec<KeyValuePair>,
+}
+
+impl<'a> Endpoint<(), (), Vec<KeyValuePair>> for WriteBulk<'a> {
+    fn method(&self) -> Method {
+        Method::Put
+    }
+    fn path(&self) -> String {
+        format!(
+            "accounts/{}/storage/kv/namespaces/{}/bulk",
+            self.account_identifier, self.namespace_identifier
+        )
+    }
+    fn body(&self) -> Option<Vec<KeyValuePair>> {
+        Some(self.bulk_key_value_pairs.clone())
+    }
+    // default content-type is already application/json
+}
+
+#[derive(Serialize, Deserialize, Clone, Debug)]
+pub struct KeyValuePair {
+    pub key: String,
+    pub value: String,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub expiration: Option<i32>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub expiration_ttl: Option<i32>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub base64: Option<bool>,
+}
--- a/src/endpoints/workerskv/mod.rs
+++ b/src/endpoints/workerskv/mod.rs
@@ -7,6 +7,7 @@ pub mod list_namespace_keys;
 pub mod list_namespaces;
 pub mod remove_namespace;
 pub mod rename_namespace;
+pub mod write_bulk;
 
 /// Workers KV Namespace
 /// A Namespace is a collection of key-value pairs stored in Workers KV.


src/workerskv/write_bulk.rs
```diff
@@ -0,0 +1,44 @@
...
+    // default content-type is already application/json
+}
+
+#[derive(Serialize, Clone, Debug)]
+pub struct BulkWriteParams<'a> {
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub expiration: Option<&'a str>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub expiration_ttl: Option<&'a str>,
+    pub base64: bool,
```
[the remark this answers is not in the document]
gabbifish: It's being added! API change PR still waiting for merge :package:

src/workerskv/write_bulk.rs
```diff
@@ -0,0 +1,44 @@
...
+    fn query(&self) -> Option<BulkWriteParams<'a>> {
+        Some(self.params.clone())
+    }
+    // default content-type is already application/json
+}
+
+#[derive(Serialize, Clone, Debug)]
+pub struct BulkWriteParams<'a> {
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub expiration: Option<&'a str>,
```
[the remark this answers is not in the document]
gabbifish: The way I understood the query params is that they were used as url params. I followed the use of Option because I wanted to make sure these params weren't included if their value was none. I was also trying to follow the example in dns.rs:
```
#[derive(Serialize, Clone, Debug, Default)]
pub struct ListDnsRecordsParams {
    #[serde(skip_serializing_if = "Option::is_none")]
    pub record_type: Option<DnsContent>,
    #[serde(skip_serializing_if = "Option::is_none")]
    pub name: Option<String>,
    #[serde(skip_serializing_if = "Option::is_none")]
    pub page: Option<u32>,
    #[serde(skip_serializing_if = "Option::is_none")]
    pub per_page: Option<u32>,
    #[serde(skip_serializing_if = "Option::is_none")]
    pub order: Option<ListDnsRecordsOrder>,
    #[serde(skip_serializing_if = "Option::is_none")]
    pub direction: Option<OrderDirection>,
    #[serde(rename = "match", skip_serializing_if = "Option::is_none")]
    pub search_match: Option<SearchMatch>,
}
```
Let me know if I misunderstood how the query params are used--I think I did it the right way, but I am also new to this code base!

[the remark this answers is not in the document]
adamchalmers: Oh oops, I'm sorry, I meant to write "should these be Option<uint>", but forgot the uint part. We should make sure users can't put "asdf" for the expiration, is all. Not sure which uint would be appropriate, but the [DNS endpoints use u32](https://github.com/cloudflare/cloudflare-rs/blob/master/src/endpoints/dns.rs#L52)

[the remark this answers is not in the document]
gabbifish: AFAIK Time since unix epoch is stored in i32, so we should probably be consistent! (Even though it really should be u32 imho)

src/endpoints/workerskv/write_bulk.rs
```diff
@@ -0,0 +1,40 @@
...
+#[derive(Serialize, Deserialize, Clone, Debug)]
+pub struct KeyValuePair {
+    pub key: String,
+    pub value: String,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub expiration: Option<i32>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub expiration_ttl: Option<i32>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub base64: Option<bool>,
```
sssilver: We should be leveraging mechanisms discussed in https://github.com/dtolnay/request-for-implementation/issues/18 for `skip_serializing_if` and `Option`

src/endpoints/workerskv/write_bulk.rs
```diff
@@ -0,0 +1,39 @@
...
+}
+
+#[derive(Serialize, Deserialize, Clone, Debug)]
+pub struct KeyValuePair {
+    pub key: String,
+    pub value: String,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub expiration: Option<i32>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub expiration_ttl: Option<i32>,
```
sssilver: In other places (`CreateDnsRecordParams`) TTL field has been defined as `u32`, which seems more meaningful considering a negative TTL doesn't make sense.

src/endpoints/workerskv/write_bulk.rs
```diff
@@ -0,0 +1,39 @@
...
+    }
+    // default content-type is already application/json
+}
+
+#[derive(Serialize, Deserialize, Clone, Debug)]
+pub struct KeyValuePair {
+    pub key: String,
+    pub value: String,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub expiration: Option<i32>,
```
sssilver: What does this hold?

src/endpoints/workerskv/write_bulk.rs
adamchalmers: see discussion [here](https://github.com/cloudflare/cloudflare-rs/pull/28#pullrequestreview-275750400)

gabbifish: I chatted with the KV team and apparently they use i64! I will use that instead (its max value is much larger than u32).

src/endpoints/workerskv/write_bulk.rs
gabbifish: It holds an expiration date (seconds since epoch)

src/endpoints/workerskv/write_bulk.rs
gabbifish: This is *borat voice* VERY NaIIICE

PR #29, opened 2019-08-19
Add support for Workers KV individual and bulk deletion API endpoints

gabbifish:
This PR adds support for the Workers KV [individual delete](https://api.cloudflare.com/#workers-kv-namespace-delete-key-value-pair) and [bulk deletion](https://api.cloudflare.com/#workers-kv-namespace-delete-multiple-key-value-pairs) API endpoints. It is used by https://github.com/cloudflare/wrangler/pull/452


src/endpoints/workerskv/delete_key.rs
```diff
@@ -0,0 +1,24 @@
...
+
+impl<'a> Endpoint<(), (), ()> for DeleteKey<'a> {
+    fn method(&self) -> Method {
+        Method::Delete
+    }
+    fn path(&self) -> String {
+        // Need multiple cases for presence of ttl or expiration parameters.
+        format!(
+            "accounts/{}/storage/kv/namespaces/{}/values/{}",
+            self.account_identifier, self.namespace_identifier, self.key
```
ashleymichal: url encode the key here (and then probably not in wrangler)

src/endpoints/workerskv/delete_bulk.rs
```diff
@@ -0,0 +1,28 @@
...
+    pub namespace_identifier: &'a str,
+    pub bulk_keys: Vec<String>,
+}
+
+impl<'a> Endpoint<(), (), Vec<String>> for DeleteBulk<'a> {
+    fn method(&self) -> Method {
+        Method::Delete
+    }
+    fn path(&self) -> String {
+        // Need multiple cases for presence of ttl or expiration parameters.
```
adamchalmers: I don't understand this comment?

ashleymichal: same

gabbifish: Oop! This is an old comment. Removing.

src/endpoints/workerskv/delete_key.rs
gabbifish: Done!

src/endpoints/workerskv/mod.rs
```diff
@@ -30,3 +33,7 @@ pub struct Key {
 
 impl ApiResult for Key {}
 impl ApiResult for Vec<Key> {}
+
+pub fn url_encode_key(key: &str) -> String {
```
sssilver: Is `pub` necessary here?

gabbifish: great catch, I don't think so--this is only used in this module :)

gabbifish:
```suggestion
fn url_encode_key(key: &str) -> String {
```

PR #30, opened 2019-08-20
Fix optional param passing for listing Workers KV namespace keys

gabbifish:
This is a small PR to ensure that all the options passed into the ListNamespaceKeys endpoint are not serialized if they are not included.


gabbifish:
> [dtolnay/request-for-implementation#18](https://github.com/dtolnay/request-for-implementation/issues/18)

this has been fixed!

PR #33, opened 2019-08-26
fix #25: Add API token support

### LINKED ISSUE #25: Add support for API Tokens
The API now supports the use of API tokens as credentials (see https://api.cloudflare.com/#getting-started-requests). We should provide this option in the client.

gabbifish:
This PR adds support for api token usage (https://api.cloudflare.com/#getting-started-requests)

### DIFF
--- a/example/main.rs
+++ b/example/main.rs
@@ -175,22 +175,18 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
     let key = matches.value_of("auth-key");
     let token = matches.value_of("auth-token");
 
-    // Create bogus default api_client; this will be overwritten with a client
-    // that uses a key or token.
-    let mut credentials = Credentials::Default;
-
-    if let Some(key) = key {
-        credentials = Credentials::UserAuthKey {
+    let credentials: Credentials = if let Some(key) = key {
+        Credentials::UserAuthKey {
             email: email.unwrap().to_string(),
             key: key.to_string(),
-        };
-    }
-
-    if let Some(token) = token {
-        credentials = Credentials::UserAuthToken {
+        }
+    } else if let Some(token) = token {
+        Credentials::UserAuthToken {
             token: token.to_string(),
-        };
-    }
+        }
+    } else {
+        panic!("Either API token or API key + email pair must be provided")
+    };
 
     let api_client = HttpApiClient::new(credentials);
 
--- a/example/main.rs
+++ b/example/main.rs
@@ -3,7 +3,7 @@ extern crate maplit;
 extern crate clap;
 extern crate cloudflare;
 
-use clap::{App, AppSettings, Arg, ArgMatches, SubCommand};
+use clap::{App, AppSettings, Arg, ArgGroup, ArgMatches, SubCommand};
 use cloudflare::endpoints::{dns, zone};
 use cloudflare::framework::{
     apiclient::ApiClient,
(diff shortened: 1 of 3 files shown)


gabbifish:
I updated the PR with the suggestions above. Please let me know what you think about the default option on the enum!

src/framework/auth.rs
```diff
@@ -13,10 +15,14 @@ pub trait AuthClient {
...
     fn auth(self, credentials: &Credentials) -> RequestBuilder {
         match credentials {
-            Credentials::User { key, email } => self
-                .header("X-Auth-Key", key.as_str())
-                .header("X-Auth-Email", email.as_str()),
+            Credentials::UserAuthKey { email, key } => self
+                .header("X-Auth-Email", email.as_str())
+                .header("X-Auth-Key", key.clone()),
+            Credentials::UserAuthToken { email, token } => self
+                .header("X-Auth-Email", email.as_str())
```
ashleymichal: according to the API docs, tokens are used by themselves, not in tandem with email

example/main.rs
```diff
@@ -165,13 +171,28 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+    let key = matches.value_of("auth-key");
+    let token = matches.value_of("auth-token");
 
-    let api_client = HttpApiClient::new(Credentials::User {
-        key: key.to_string(),
-        email: email.to_string(),
-    });
+    // Create bogus default api_client; this will be overwritten with a client
+    // that uses a key or token.
+    let mut credentials = Credentials::Default;
```
adamchalmers: I don't think we need to add a special `Default` variant just for this assignment. Instead, we can assign `credentials` to the result of an `if`, like this:

```rust
let credentials = if let Some(key) = key {
    Credentials::UserAuthKey {
        email: email.unwrap().to_string(),
        key: key.to_string(),
    }
} else if let Some(token) = token {
    Credentials::UserAuthToken {
        token: token.to_string(),
    }
} else {
    panic!("either token or key must be specified")
}
```

src/framework/auth.rs
```diff
@@ -2,8 +2,10 @@ use reqwest::RequestBuilder;
 
 #[derive(Debug)]
 pub enum Credentials {
-    User { key: String, email: String },
+    UserAuthKey { email: String, key: String },
+    UserAuthToken { token: String },
     Service { key: String },
+    Default,
```
adamchalmers: See my comment on `example/main.rs` - I don't think we should add a `Default` variant. We don't ever want a user to actually use it, so let's just never implement it in the first place :)

src/framework/auth.rs
```diff
@@ -13,10 +15,14 @@ pub trait AuthClient {
...
-                .header("X-Auth-Key", key.as_str())
-                .header("X-Auth-Email", email.as_str()),
+            Credentials::UserAuthKey { email, key } => self
+                .header("X-Auth-Email", email.as_str())
+                .header("X-Auth-Key", key.clone()),
+            Credentials::UserAuthToken { token } => {
+                self.header("Authorization", &format!("Bearer {}", token.clone()))
+            }
             Credentials::Service { key } => self.header("X-Auth-User-Service-Key", key.as_str()),
+            _ => self,
```
adamchalmers: If you remove the `Default` variant, don't forget to remove this too.

example/main.rs
gabbifish: Ah! I'll try this. I tried using other error handling approaches (namely, failure crate) but the compiler was unhappy with me; just using `panic!` should work though!

src/framework/auth.rs
gabbifish: Agreed, it was more of a perceived necessary evil at the time :)

example/main.rs
adamchalmers: Great. I'd never put a `panic!` in the library itself, but in the example binary it's totally fine.

PR #34, opened 2019-09-09
Need to parse KV API times as i64

gabbifish:
This is a small bug fix; the KV API returns i64 values for time, not UTC timestamps. Would love to get this in before we release our upcoming cloudflare-rs version.

### DIFF
--- a/src/endpoints/workerskv/mod.rs
+++ b/src/endpoints/workerskv/mod.rs
@@ -1,5 +1,9 @@
 use crate::framework::response::ApiResult;
+use chrono::DateTime;
+use chrono::{TimeZone, Utc};
 use percent_encoding::{percent_encode, PATH_SEGMENT_ENCODE_SET};
+
+use serde::{Deserialize, Deserializer};
 
 pub mod create_namespace;
 pub mod delete_bulk;
@@ -28,7 +32,23 @@ impl ApiResult for Vec<WorkersKvNamespace> {}
 #[derive(Deserialize, Serialize, Debug, Clone, PartialEq, Eq)]
 pub struct Key {
     pub name: String,
-    pub expiration: Option<i64>,
+    #[serde(default)]
+    #[serde(deserialize_with = "deserialize_option_timestamp")]
+    pub expiration: Option<DateTime<Utc>>,
+}
+
+pub fn deserialize_option_timestamp<'de, D>(
+    deserializer: D,
+) -> Result<Option<DateTime<Utc>>, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    let s: Option<i64> = Option::deserialize(deserializer)?;
+    if let Some(s) = s {
+        return Ok(Some(Utc.timestamp(s, 0)));
+    }
+
+    Ok(None)
 }
 
 impl ApiResult for Key {}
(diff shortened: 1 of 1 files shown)


sssilver:
I'm not entirely sure we shouldn't be parsing those i64s into timestamps.

gabbifish:
@sssilver if I don't parse the api values into i64, I get errors like `Error: invalid type: integer `1568014659`, expected a formatted date and time string or a unix timestamp at line 5 column 30`--I don't think there is any automatic functionality for converting these api-returned values into UTC timestamps :/ 

tl;dr I don't really see a way around this, but if you have ideas, I'd love to know them!

adamchalmers:
I see Areg's point - the API returns i64 because JSON doesn't have a type for time. But as this is the Rust API client, perhaps we should use Serde's [deserialize_with](https://serde.rs/field-attrs.html#deserialize_with) field attribute to parse the i64 in the JSON into a Rust time object. 

Pros:
* it's more idiomatic Rust to store time as an actual `chrono` struct rather than some numeric type
* it'd be what the API did if JSON supported times

Cons:
* it's tampering with the actual API response, some users might just want the actual API response
* if users want a `chrono` object they can easily instantiate one themselves, maybe we shouldn't assume that's what the user wants

gabbifish:
@adamchalmers Agreed with Areg's input :) I now use deserialize_with. I think it actually is better for users to see timestamps instead of the epoch seconds the API returns; it is certainly more usable for them! As for users wanting a chrono object, I think we (as the only users of this kv endpoint module) are happy to make this a default in cloudflare-rs.

src/endpoints/workerskv/mod.rs
```diff
@@ -30,9 +32,25 @@ impl ApiResult for Vec<WorkersKvNamespace> {}
...
 pub struct Key {
     pub name: String,
+    #[serde(default)]
+    #[serde(deserialize_with = "deserialize_option_timestamp")]
     pub expiration: Option<DateTime<Utc>>,
 }
 
+pub fn deserialize_option_timestamp<'de, D>(
+    deserializer: D,
+) -> Result<Option<DateTime<Utc>>, D::Error>
```
adamchalmers: I wonder if this return type should be `Result<DateTime<Utc>>, D::Error>` instead. You could argue that this function should only return `Ok` if the JSON response is deserialized into an actual DateTime<Utc>. 

From the perspective of the application code using `cloudflare-rs`, if I call an endpoint that's supposed to contain a timestamp, and it doesn't return an actual timestamp, it's a `Err` not `Ok`. So maybe instead of returning `Ok(None)` it should return `Err`. What do you think, as the person who's actually using this endpoint?

sssilver: Awesome PR!

If I understand him correctly, I agree with @adamchalmers. It feels like the deserializer should be agnostic of whether the value it's deserializing is wrapped in an optional. I was therefore expecting more of a `deserialize_timestamp` rather than a `deserialize_option_timestamp`. Am I missing a nuance here?

gabbifish: Since `expiration` an optional value, I think it still makes sense to return an Option? Not having `None` would make the semantics of using the `Key` struct intuitive, and also would not play well with the serde optional deserialization logic we're using.

adamchalmers: Oh, I didn't realize that `expiration` is defined as optional in the API. Nevermind then, I agree with your implementation.

PR #46, opened 2019-10-02
Add support for custom config for clients

### LINKED ISSUE #45: Allow users to pass in a timeout for the API client
This is related to https://github.com/cloudflare/wrangler/issues/746.

Ideally, users should be allowed to specify the API client timeout period when constructing the API client with `HttpApiClient::new()`. What do you think?

gabbifish:
This PR is a POC for custom config support with the cloudflare-rs API client.

So far, it only implements timeouts to close #45. This approach can be extended to more arguments, however.

This design was pulled from a conversation with @sssilver, in which he suggested passing a new struct that contains config information. This struct is passed in a `new_custom` constructor as to avoid breaking callers of the preexisting `new()` constructor. Please let me know if this is what you expected!


adamchalmers:
There's a lot of other commits in this PR, do you need to rebase on origin/master?

src/framework/mod.rs
```diff
@@ -56,6 +62,23 @@ impl HttpApiClient {
...
     }
+
+    pub fn new_custom(
+        credentials: auth::Credentials,
+        config: HttpApiClientConfig,
+    ) -> HttpApiClient {
+        let http_client = reqwest::Client::builder()
+            .timeout(config.http_timeout)
+            .build()
+            .unwrap(); // Ok to unwrap; the only time this fails is if the native TLS backend cannot
```
adamchalmers: If this value is `Err`, we've basically got a choice about how to handle - unwrap immediately, or do something (e.g. print a nice error). If we use `unwrap()`, we're making that choice for the user. I suggest changing this to `?` and making the method return `Result/Option` instead. Because that way, if the user just wants to `unwrap()` they're free to do so. But if they want to handle it some other way, they can do that too. We should give the user the flexibility to do that :)

gabbifish: True! My initial thought about using unwrap() was that, if we return a `Result`, we will diverge from the usage of `new()` above--how do we feel about that?

adamchalmers: I feel fine - if the programmer doesn't want to handle it, they're free to unwrap it. Optionals/results are so common in the stdlib and this crate that I feel anyone using this crate will need to be familiar with them anyway.

gabbifish: Cool! Using Result now :)

src/framework/mod.rs
```diff
@@ -48,6 +50,10 @@ pub struct HttpApiClient {
     http_client: reqwest::Client,
 }
 
+pub struct HttpApiClientConfig {
```
sssilver: I would say instead of making every field an `Option`, perhaps we should use https://doc.rust-lang.org/std/default/trait.Default.html

src/framework/mod.rs
```diff
@@ -56,6 +62,21 @@ impl HttpApiClient {
             http_client: reqwest::Client::new(),
         }
     }
+
+    pub fn new_custom(
```
sssilver: Let's just call this `new()`. People can call `HttpApiClientConfig::default()` if they just want to pass the default values.

adamchalmers: Breaking change though :/

ashleymichal: #versioning

adamchalmers: but I guess we have so few external users right now that it's OK

PR #49, opened 2019-10-23
Document the SearchMatch type

adamchalmers:


src/framework/mod.rs
```diff
@@ -22,10 +22,14 @@ pub enum OrderDirection {
     Descending,
 }
 
+/// Used as a parameter to API calls that search for a resource (e.g. DNS records).
+/// Tells the API whether to return results that match all search requirements or at least one (any).
```
ashleymichal: is this as general as it sounds? like, is this a standard pattern across cloudflare apis? or would it make more sense in the dns endpoint library? This is a genuine question, most of my experience is with workers ecosystem apis.

adamchalmers: I'm not sure if it's used anywhere besides DNS, but it's a general-purpose idea... any other endpoint could use it, I guess.

PR #51, opened 2019-10-23
fix #50: ApiFailure and ApiError impl std::error::Error

### LINKED ISSUE #50: ApiFailure should impl Error
To integrate with Failure and the rest of the Rust error ecosystem, `std::error::Error` should be implemented for `cloudflare::framework::response::apifail::ApiFailure`. Otherwise, it's hard to use `cloudflare-rs` with `?` in a function that returns `Fallible`.

adamchalmers:


src/framework/response/apifail.rs
```diff
@@ -37,6 +38,7 @@ impl PartialEq for ApiError {
 
 impl Eq for ApiError {}
 impl Eq for ApiErrors {}
+impl Error for ApiError {}
```
nickvollmar: Not ApiErrors?

Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "cloudflare"
-version = "0.3.1"
+version = "0.4.1"
```
sssilver: Isn't this something we only bump up during release?

src/framework/response/apifail.rs
adamchalmers: No, ApiErrors doesn't impl Display, which is required for impling Error.

Cargo.toml
adamchalmers: Good point

PR #53, opened 2019-10-23
Fix #52: Allow custom dev environments

### LINKED ISSUE #52: Add dev environments (with custom URLs)
Currently Cloudflare-rs only supports production environments. But users need to test their code against other environments too, e.g. a local developer instance of the Cloudflare API. Users should be able to provide a custom URL.

adamchalmers:


src/framework/mod.rs
```diff
@@ -36,6 +36,7 @@ pub enum SearchMatch {
 #[derive(Debug)]
 pub enum Environment {
     Production,
+    Dev(url::Url),
```
sssilver: Not sure if `Dev` is better or `Custom` is better.

adamchalmers: I thought Dev was better, but now I actually prefer Custom, for no real reason.

PR #55, opened 2019-10-24
Add user detail endpoint

gabbifish:
This PR adds support for the user-detail endpoint, so a user can get information (e.g. email) about themselves via the API :)


src/endpoints/user.rs
```diff
@@ -0,0 +1,36 @@
...
+    pub first_name: String,
+    pub last_name: String,
+    pub username: String,
+    pub telephone: String,
+    pub contry: String,
+    pub zipcode: String,
+    pub created_on: DateTime<Utc>,
+    pub modified_on: DateTime<Utc>,
+    pub two_factor_authentication_enabled: bool,
+    pub suspented: bool,
```
adamchalmers: typo: should be "suspended"

gabbifish: omfg thank u <3

gabbifish:
```suggestion
    pub suspended: bool,
```

PR #58, opened 2019-11-05
Add missing fields from user endpoint

gabbifish:
This PR adds additional fields that are present in /user endpoint responses.

### DIFF
--- a/src/endpoints/user.rs
+++ b/src/endpoints/user.rs
@@ -11,9 +11,9 @@ use chrono::{DateTime, Utc};
 pub struct Organization {
     id: String,
     name: String,
-    status: String,
-    permissions: Vec<String>,
-    roles: Vec<String>,
+    status: String, // Whether or not the user is a member of the organization or has an inivitation pending
+    permissions: Vec<String>, // Access permissions for this User
+    roles: Vec<String>, // List of role names for the User at the Organization
 }
 
 #[derive(Deserialize, Serialize, Debug, Clone, PartialEq, Eq)]
--- a/src/endpoints/user.rs
+++ b/src/endpoints/user.rs
@@ -8,19 +8,30 @@ use chrono::{DateTime, Utc};
 /// https://api.cloudflare.com/#user-user-details
 
 #[derive(Deserialize, Serialize, Debug, Clone, PartialEq, Eq)]
+pub struct Organization {
+    id: String,
+    name: String,
+    status: String,
+    permissions: Vec<String>,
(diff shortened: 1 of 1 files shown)


src/endpoints/user.rs
```diff
@@ -7,20 +7,31 @@ use chrono::{DateTime, Utc};
 /// Gets information about a user
 /// https://api.cloudflare.com/#user-user-details
 
+#[derive(Deserialize, Serialize, Debug, Clone, PartialEq, Eq)]
+pub struct Organization {
```
adamchalmers: It could be useful to add a description of these fields - you can just copy and paste the descriptions from https://api.cloudflare.com/#user-properties (click organizations > show definition)

src/endpoints/user.rs
```diff
@@ -7,20 +7,31 @@ use chrono::{DateTime, Utc};
...
+    pub telephone: Option<String>,
+    pub zipcode: Option<String>,
+    pub last_name: Option<String>,
+    pub modified_on: DateTime<Utc>,
     pub username: String,
-    pub telephone: String,
-    pub contry: String,
-    pub zipcode: String,
     pub created_on: DateTime<Utc>,
-    pub modified_on: DateTime<Utc>,
```
adamchalmers: The API docs say there's still a `modified_on` field, are they wrong or is there some other reason for deleting them?

src/endpoints/user.rs
```diff
@@ -7,20 +7,31 @@ use chrono::{DateTime, Utc};
 /// Gets information about a user
 /// https://api.cloudflare.com/#user-user-details
 
+#[derive(Deserialize, Serialize, Debug, Clone, PartialEq, Eq)]
+pub struct Organization {
+    id: String,
+    name: String,
+    status: String, // Whether or not the user is a member of the organization or has an inivitation pending
+    permissions: Vec<String>, // Access permissions for this User
```
sssilver: Feels weird that permissions is a vector of Strings. Should probably be a vector of Permissions.

ashleymichal: in terms of like an Enum called Permissions? I think I agree, but what is tricky about that is the list is growing and not under our control. We'd have to keep track of the existing list of permissions managed by the API team and keep in lockstep to prevent breakage. Not 100% opposed, but just pointing out the additional overhead of maintaining that enum.

ashleymichal: it also feels weird that these permissions are part of the organization struct but i think i just don't get this api.

PR #60, opened 2019-11-06
Alewis/suggested user changes

ashleymichal:
Per comments on #59 ->

* match directory structure used in WorkersKV
* remove redundant `User` qualifiers from `user` module structs


src/endpoints/user/mod.rs
```diff
@@ -17,7 +17,7 @@ pub struct Organization {
 }
 
 #[derive(Deserialize, Serialize, Debug, Clone, PartialEq, Eq)]
-pub struct UserDetails {
+pub struct Details {
```
adamchalmers: Is this worth the breaking change? 
EDIT: Actually I can see that [Wrangler is the only crate on crates.io that uses this](https://crates.io/crates/cloudflare/reverse_dependencies) so sure, let's do this stuff while we can.

ashleymichal: still a good point, as if the struct existed before @gabbifish and I started mucking about in here, chances are tunnel team is using it and will have to adapt also.

sssilver: Also, does this change actually improve anything? Is my understanding correct that the motivation for this is that `user::Details` is sufficiently descriptive? If so, then I somewhat disagree with the idea. "Details" is a generic enough word that within the context of "user" it may be confusing what details it's referring to -- user details? token details? <type defined in module user> details?

sssilver: I think in general it'll be more intuitive for the API user for terms to match general Cloudflare API docs one-to-one whenever possible.

ashleymichal: happy for this PR to be rejected or split :D, not super opinionated on this point.

gabbifish:
We will hold off on refactoring for now.

PR #64, opened 2019-11-08
Fix #61, 63: async API client

### LINKED ISSUE #61: Add async ApiClient
Now that async/await is in stable Rust 1.39, we should support async. I know Tunnelmon would be much cheaper to run if we used an event loop instead of parallelizing across cores.

I'm not sure it's possible to use the same ApiClient trait for both sync and async (because one will return Future<T> and one will return T). So we might need a second trait.

adamchalmers:
Features:

- an async API client under `framework::async_api::Client` that works just like `framework::apiclient::ApiClient`, except there's no trait for it (because there's no async fn in traits yet)
- e2e tests updated to use the async client, to prove it works.
- updated some other dependencies

Future work:

You'll notice that `framework/async_api.rs` has some repeated code from `framework/mod.rs` Unfortunately, `reqwest::{request, client, response}` and `reqwest::blocking::{request, client, response}` don't share any traits, so it's not clear how best to share code across the a/sync implementations. I've factored out as much as I can with plain old functions. Please let me know if you have any suggestions for making the code more DRY!

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,14 +1,26 @@
 # Contributing to cloudflare-rs
 
-There are two root-level modules in `cloudflare-rs`. Most PRs will only touch one of them. If you're
-working on the API framework itself, all the relevant code lives under the `framework/`
-module. If you're adding or editing a particular endpoint, read the "Adding New Endpoints" section below.
+There are two root-level modules in `cloudflare-rs`. Most PRs will only touch one of them. The
+`framework/` module contains the API framework itself. The `endpoints/` module contains code for
+each Cloudflare API endpoint, grouped by product, e.g. `endpoints/dns` or `endpoints/zones`. Check 
+the "Updating the Framework" or "Adding New Endpoints" sections below.
 
 ## Pull Requests
 
-Every PR should have a corresponding issue, and the issue number should appear in the PR's description and commit message.
+Every PR should have a corresponding issue, and the issue number should appear in the PR's 
+description and commit message.
 
 PRs should be squashed to one commit before merging.
+
+## Updating the Framework
+
+This library includes both async and blocking API clients. The `ApiClient` trait covers blocking 
+requests, and is implemented by the `HttpApiClient` struct. The `async_api::Client` struct covers
+async requests. Because Rust doesn't support async fns in traits yet, there is no trait for async
+API clients.
+
+If you want to change how the Cloudflare API client works, please remember to make the change in 
+both blocking and async clients if applicable.
 
 ## Adding New Endpoints
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,20 +21,19 @@ name = "e2e-tests"
 path = "e2e_tests/main.rs"
 
 [dependencies]
+tokio = { version = "0.2", features = ["macros"] } # Only used in bin/e2e, not used in library.
 chrono = { version = "0.4", features = ["serde"] }
-clap = "2.33"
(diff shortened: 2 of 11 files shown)


gabbifish:
Hi! Your change to fix the action logic (to install a specific cargo version) fixed the CI build fails in this PR. I hope you don't mind me merging master into this branch for ya :) Now this shall be ready for review, which I'll get to on Monday!

adamchalmers:
Thank you Gabbi!

adamchalmers:
Upgraded to reqwest 0.10 (previously was 0.10-alpha). Can someone please review this so we can merge?

Cargo.toml
```diff
@@ -21,20 +21,19 @@ name = "e2e-tests"
 path = "e2e_tests/main.rs"
 
 [dependencies]
+tokio = { version = "0.2", features = ["macros"] } # Only used in bin/e2e, not used in library.
```
EverlastingBugstopper: is `bin/e2e` the same as the `binaries` comment on line 26? if so i'd nit them to be consistent

src/lib.rs
```diff
@@ -1,9 +1,8 @@
 ///! An API client for the [Cloudflare API](https://api.cloudflare.com)
 extern crate chrono;
 extern crate reqwest;
-extern crate serde;
 #[macro_use]
-extern crate serde_derive;
```
EverlastingBugstopper: This is interesting - is there a reason this is no longer needed?

adamchalmers: Serde now has a feature which re-exports the `derive` macros from this crate. I figured it was cleaner to just depend on the serde crate rather than both serde and serde-derive.

Cargo.toml
adamchalmers: Nah, multiple binaries (e.g. both bin/e2e and bin/example) depend on `clap`. But only one binary uses `tokio`.

Cargo.toml
```diff
@@ -21,20 +21,19 @@ name = "e2e-tests"
...
 chrono = { version = "0.4", features = ["serde"] }
-clap = "2.33"
+clap = "2.33" # Only used in binaries, not used in library.
```
sssilver: This makes me realize that we may be packaging this wrong

adamchalmers: We should probably use [cargo workspaces](https://doc.rust-lang.org/book/ch14-03-cargo-workspaces.html) instead

ashleymichal: should we make an issue?

adamchalmers: https://github.com/cloudflare/cloudflare-rs/issues/77

CONTRIBUTING.md
```diff
@@ -1,15 +1,27 @@
...
 
+## Updating the Framework
+
+This library includes both async and blocking API clients. The `ApiClient` trait covers blocking 
+requests, and is implemented by the `HttpApiClient` struct. The `async_api::Client` struct covers
+async requests. Because Rust doesn't support async fns in traits yet, there is no trait for async
+API clients.
+
+If you want to change how the Cloudflare API client works, please remember to make the change in 
+both blocking and async clients if applicable.
```
ashleymichal: 👍

src/framework/auth.rs
```diff
@@ -7,20 +7,29 @@ pub enum Credentials {
     Service { key: String },
 }
 
+impl Credentials {
+    pub fn headers(&self) -> Vec<(&'static str, String)> {
```
ashleymichal: i like this refactor. v. nice

PR #69, opened 2019-11-21
Alewis/fix#68

ashleymichal:
I can split things out if it makes sense; this PR adds support for Workers Routes endpoints, specifically Create Route, Delete Route, and List Routes, for use in Wrangler.

It also introduces an alternate print_response function in example/main.rs for outputting json results.

### DIFF
--- a/example/main.rs
+++ b/example/main.rs
@@ -12,6 +12,7 @@ use cloudflare::framework::{
     response::{ApiFailure, ApiResponse, ApiResult},
     Environment, HttpApiClient, HttpApiClientConfig, OrderDirection,
 };
+use serde::Serialize;
 
 type SectionFunction<ApiClientType> = fn(&ArgMatches, &ApiClientType);
 
@@ -24,6 +25,31 @@ struct Section<'a, ApiClientType: ApiClient> {
 fn print_response<T: ApiResult>(response: ApiResponse<T>) {
     match response {
         Ok(success) => println!("Success: {:#?}", success),
+        Err(e) => match e {
+            ApiFailure::Error(status, errors) => {
+                println!("HTTP {}:", status);
+                for err in errors.errors {
+                    println!("Error {}: {}", err.code, err.message);
+                    for (k, v) in err.other {
+                        println!("{}: {}", k, v);
+                    }
+                }
+                for (k, v) in errors.other {
+                    println!("{}: {}", k, v);
+                }
+            }
+            ApiFailure::Invalid(reqwest_err) => println!("Error: {}", reqwest_err),
+        },
+    }
+}
+
+/// Sometimes you want to pipe results to jq etc
+fn print_response_json<T: ApiResult>(response: ApiResponse<T>)
+where
+    T: Serialize,
+{
+    match response {
+        Ok(success) => println!("{}", serde_json::to_string(&success.result).unwrap()),
         Err(e) => match e {
             ApiFailure::Error(status, errors) => {
                 println!("HTTP {}:", status);
--- a/example/main.rs
+++ b/example/main.rs
@@ -170,7 +170,9 @@ fn delete_route<ApiClientType: ApiClient>(arg_matches: &ArgMatches, api_client: 
         .expect(&zone_id_missing);
 
     let route_id_missing = format!("missing '{}': {}", "ROUTE_PATTERN", usage);
-    let route_id = arg_matches.value_of("route_id").expect(&route_id_missing);
+    let route_id = arg_matches
+        .value_of("route_identifier")
+        .expect(&route_id_missing);
 
     let response = api_client.request(&workers::DeleteRoute {
         zone_identifier,
--- a/example/main.rs
+++ b/example/main.rs
@@ -4,7 +4,7 @@ extern crate clap;
 extern crate cloudflare;
 
 use clap::{App, AppSettings, Arg, ArgMatches, SubCommand};
-use cloudflare::endpoints::{dns, zone};
+use cloudflare::endpoints::{dns, workers, zone};
 use cloudflare::framework::{
     apiclient::ApiClient,
     auth::Credentials,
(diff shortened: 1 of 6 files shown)


src/endpoints/workers/create_route.rs
```diff
@@ -0,0 +1,29 @@
...
+    fn path(&self) -> String {
+        format!("zones/{}/workers/routes", self.zone_identifier)
+    }
+    fn body(&self) -> Option<CreateRouteParams> {
+        Some(self.params.clone())
+    }
+}
+
+#[derive(Serialize, Clone, Debug)]
+pub struct CreateRouteParams {
```
adamchalmers: Could you please copy the docs for these params from the API documentation (https://api.cloudflare.com/#worker-routes-create-route)? They'd be good docstrings.

adamchalmers: This could perhaps just be called `Params` so people can do `create_route::Params` instead of `create_route::CreateRouteParams`.

ashleymichal: can do

ashleymichal: i'm into it; but following the pattern established by existing endpoints

example/main.rs
```diff
@@ -42,6 +43,31 @@ fn print_response<T: ApiResult>(response: ApiResponse<T>) {
     }
 }
 
+/// Sometimes you want to pipe results to jq etc
```
gabbifish: Can we also print out errors in JSON format? Right now, this function's output can only be piped out to `jq` if the response is a success. The error messages from the API should also be json-compatible.

example/main.rs
```diff
@@ -96,6 +122,64 @@ fn create_txt_record<ApiClientType: ApiClient>(
...
+fn delete_route<ApiClientType: ApiClient>(arg_matches: &ArgMatches, api_client: &ApiClientType) {
+    let usage = "usage: delete_route ZONE_ID ROUTE_ID";
+
+    let zone_id_missing = format!("missing '{}': {}", "ZONE_ID", usage);
+    let zone_identifier = arg_matches
+        .value_of("zone_identifier")
+        .expect(&zone_id_missing);
+
+    let route_id_missing = format!("missing '{}': {}", "ROUTE_PATTERN", usage);
+    let route_id = arg_matches.value_of("route_id").expect(&route_id_missing);
```
gabbifish: Does this line actually work? The registered clap arguments below seem to refer to a `route_identifier` instead of a `route_id`

src/endpoints/workers/create_route.rs
gabbifish: it may be worth making a separate issue for addressing repo-wide stutter in struct names.

src/endpoints/workers/delete_route.rs
```diff
@@ -0,0 +1,23 @@
...
+/// Deletes a route by route id
+/// https://api.cloudflare.com/#worker-routes-delete-route
+pub struct DeleteRoute<'a> {
+    pub zone_identifier: &'a str,
+    pub identifier: &'a str,
+}
+
+impl<'a> Endpoint<WorkersRouteIdOnly> for DeleteRoute<'a> {
+    fn method(&self) -> Method {
+        Method::Post
```
gabbifish: Huh! Shouldn't the method for this be delete?

src/endpoints/workers/list_routes.rs
```diff
@@ -0,0 +1,19 @@
+use super::WorkersRoute;
+
+use crate::framework::endpoint::{Endpoint, Method};
+
+/// Create a Route
```
gabbifish: copypasta comment from route creation

example/main.rs
ashleymichal: i tried to but adding serialize impls to the API error structs was significantly more overhead, and `example/main` is for manual testing anyway, so if errors break pipes i'm not personally too bothered.

gabbifish: Huh! I would have thought serde would have offered a nice default implementation. (`#derive[default]`)

example/main.rs
ashleymichal: hrm. thought i ran a manual test for delete but it seems like i did not. will fix and re-do tests.

src/endpoints/workers/delete_route.rs
ashleymichal: i... think so. ok now i _know_ i didn't do manual tests on delete.

src/endpoints/workers/create_route.rs
ashleymichal: tbf, i actually export all public structs at the top level of the module anyway, so i import `workers::{CreateRoute, CreateRouteParams}` in wrangler for example

example/main.rs
ashleymichal: i can give it a try; i tried a derive deserialize

gabbifish: We'll shelve this for now :)

example/main.rs
```diff
@@ -96,6 +122,66 @@ fn create_txt_record<ApiClientType: ApiClient>(
...
+    let zone_identifier = arg_matches
+        .value_of("zone_identifier")
+        .expect(&zone_id_missing);
+
+    let route_pattern_missing = format!("missing '{}': {}", "ROUTE_PATTERN", usage);
+    let route_pattern = arg_matches
+        .value_of("route_pattern")
+        .expect(&route_pattern_missing);
+
+    let script_name = arg_matches.value_of("script_name");
```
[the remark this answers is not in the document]
gabbifish: oh wait, I think `.map()` expects results anyways. Ignore me!

src/endpoints/workers/mod.rs
```diff
@@ -0,0 +1,39 @@
...
+
+/// Workers KV Route
+/// Routes are basic patterns used to enable or disable workers that match requests.
+/// https://api.cloudflare.com/#worker-routes-properties
+#[derive(Deserialize, Serialize, Debug, Clone, PartialEq, Eq)]
+pub struct WorkersRoute {
+    /// Namespace identifier tag.
+    pub id: String,
+    /// The basic pattern that should map to the script
+    pub pattern: String,
```
gabbifish: Is there any interest in combining this struct and `WorkersRouteIdOnly` and using `#[serde_with::skip_serializing_none]` to only try to serialize `pattern` and `script` if those fields are present in the API response?

ashleymichal: so the reason i did this is that one endpoint only returns an id, and others return all of them; they are not "optional" exactly, there are just different structures based on the endpoint. if we do skip, i would say we should make a PR to the config api to return the whole object all the time instead of changing the entity that is returned based on the endpoint.

ashleymichal: merging this way; if the api changes it won't break the client, but we'll just miss out on those fields until we update.

PR #73, opened 2019-12-06
feat(endpoints): add account and list account

ashleygwilliams:


src/endpoints/account/list_accounts.rs
```diff
@@ -0,0 +1,30 @@
...
+        Some(self.params.clone())
+    }
+}
+
+#[serde_with::skip_serializing_none]
+#[derive(Serialize, Clone, Debug, Default)]
+pub struct ListAccountsParams {
+    pub page: Option<u32>,
+    pub per_page: Option<u32>,
+    pub direction: Option<String>
```
ashleymichal: i think documenting with a comment that this will only work with the strings `asc` and `desc` is worthwhile.

ashleygwilliams: i'm gonna do the enum haha

Issue #77, opened 2020-01-06
Use cargo workspaces


adamchalmers:
At the moment, crates which depend on `cloudflare-rs` will pull in `clap` and `tokio`, which are large crates that take a long time to download and compile. This sucks, because cloudflare-rs' library doesn't even use them! They're only used in the example/testing binaries. We shouldn't make users compile times unnecessarily long.

Solution: use cargo workspaces, and make the library a separate project from the example and the e2e tests. That way the library will only declare dependencies it actually uses.

PR #80, opened 2020-01-07
Update publish.yml to use new cargo workspace

adamchalmers:
When I did #77 the project structure changed. I forgot to update the release.yml script to use the new project structure.


.github/workflows/publish.yml
```diff
@@ -17,6 +17,6 @@ jobs:
       env:
         CRATES_IO_TOKEN: ${{ secrets.crates_io_token }} # https://help.github.com/en/articles/virtual-environments-for-github-actions#creating-and-using-secrets-encrypted-variables
     - name: Dry run publish
-      run: cargo publish --dry-run
+      run: cd cloudflare && cargo publish --dry-run
```
ashleygwilliams: does this dir change persist to the next line?

adamchalmers: Hopefully. Broke it into its own line instead.

PR #82, opened 2020-01-07
Fix warnings from a previous PR

adamchalmers:


cloudflare-examples/src/main.rs
```diff
@@ -135,10 +135,8 @@ fn list_routes<ApiClientType: ApiClient>(arg_matches: &ArgMatches, api_client: &
     print_response_json(response);
 }
 
-fn list_accounts<ApiClientType: ApiClient>(arg_matches: &ArgMatches, api_client: &ApiClientType) {
-    let usage = "usage: list_accounts";
-
-    let response = api_client.request(&account::ListAccounts { params: None,});
+fn list_accounts<ApiClientType: ApiClient>(_arg_matches: &ArgMatches, api_client: &ApiClientType) {
```
adamchalmers: This has to take a parameter because the function has to be of type `(&ArgMatches, &ApiClientType)` to match the other functions being stored in a hashmap.

PR #84, opened 2020-01-08
Fix #26: Get response as binary

### LINKED ISSUE #26: HttpApiClient should be able to handle non-JSON request and response bodies

adamchalmers:
Workers KV API endpoints return binary responses. This is a problem because our whole API structure assumes responses should be deserialized into JSON.

The simplest solution I've found so far is expanding the `ApiClient` trait. Currently it has one method, `request`. This PR adds a second method, `request_raw_bytes`. Instead of returning JSON, it returns bytes. So:

```rust
fn request          (&dyn Endpoint<ResultType, QueryType, BodyType>) -> ApiResponse<ResultType>
fn request_raw_bytes(&dyn Endpoint<ResultType, QueryType, BodyType>) -> Result<Vec<u8>, reqwest::Error>
```

This is nice, because some customers down the road might not want to deserialize the API responses immediately -- maybe they want to keep them as bytes and send them somewhere else. 

**Improvements for the future**
Currently, you can define an endpoint which should only ever be used with `request_raw_bytes` (like the Workers KV endpoints, which should never be deserialized into JSON). But the compiler won't enforce that. It'd be great if in the future, trying to pass Workers KV endpoint into `request()` is a compile error. 

I tried to implement this, but because `request` already takes a parameter which is `dyn Endpoint`, you can't apply a second trait bound for `trait JsonResponse`. Trying to do so gives you the same error as [this SO question](https://stackoverflow.com/questions/49710852/what-is-an-auto-trait-in-rust)

This is _somewhat_ mitigated by the fact that you can define the binary-only endpoint with `ResultType = ()`, which means that `request()` will return `()`. That can be a signal that the endpoint shouldn't actually deserialize into anything.

### DIFF
--- a/cloudflare/src/framework/apiclient.rs
+++ b/cloudflare/src/framework/apiclient.rs
@@ -5,13 +5,23 @@ use crate::framework::{
 };
 use serde::Serialize;
 
-/// Synchronously sends requests to the Cloudflare API.
+/// Blocks and sends requests to the Cloudflare API.
 pub trait ApiClient {
-    /// Synchronously send a request to the Cloudflare API.
+    /// Block and send a request to the Cloudflare API, deserializing the JSON response.
     fn request<ResultType, QueryType, BodyType>(
         &self,
         endpoint: &dyn Endpoint<ResultType, QueryType, BodyType>,
     ) -> ApiResponse<ResultType>
+    where
+        ResultType: ApiResult,
+        QueryType: Serialize,
+        BodyType: Serialize;
+
+    /// Block and send a request to the Cloudflare API, get the response as bytes.
+    fn request_raw_bytes<ResultType, QueryType, BodyType>(
+        &self,
+        endpoint: &dyn Endpoint<ResultType, QueryType, BodyType>,
+    ) -> Result<Vec<u8>, reqwest::Error>
     where
         ResultType: ApiResult,
         QueryType: Serialize,
--- a/cloudflare/src/framework/async_api.rs
+++ b/cloudflare/src/framework/async_api.rs
@@ -8,15 +8,28 @@ use crate::framework::{
     Environment, HttpApiClientConfig,
 };
 use async_trait::async_trait;
+use bytes::Bytes;
 use reqwest;
 use serde::Serialize;
 
+/// Sends requests to the Cloudflare API.
 #[async_trait]
 pub trait ApiClient {
+    /// Send a request to a particular Cloudflare API endpoint, deserialize the JSON response.
     async fn request<ResultType, QueryType, BodyType>(
         &self,
         endpoint: &(dyn Endpoint<ResultType, QueryType, BodyType> + Send + Sync),
     ) -> ApiResponse<ResultType>
+    where
+        ResultType: ApiResult,
+        QueryType: Serialize,
+        BodyType: Serialize;
+
+    /// Send a request to a particular Cloudflare API endpoint, get the response as bytes.
+    async fn request_raw_bytes<ResultType, QueryType, BodyType>(
+        &self,
+        endpoint: &(dyn Endpoint<ResultType, QueryType, BodyType> + Send + Sync),
+    ) -> Result<Bytes, reqwest::Error>
     where
         ResultType: ApiResult,
         QueryType: Serialize,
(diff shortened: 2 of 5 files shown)


cloudflare/src/framework/apiclient.rs
```diff
@@ -15,4 +15,13 @@ pub trait ApiClient {
         ResultType: ApiResult,
         QueryType: Serialize,
         BodyType: Serialize;
+
+    /// Synchronously send a request to the Cloudflare API, get the response as bytes.
+    fn request_bytes<QueryType, BodyType>(
```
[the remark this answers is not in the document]
adamchalmers: Done!

ashleymichal:
so, wild idea: could an ApiResponse be an enum?

🏃‍♀ 💨

cloudflare/src/framework/async_api.rs
```diff
@@ -21,6 +24,16 @@ pub trait ApiClient {
         ResultType: ApiResult,
         QueryType: Serialize,
         BodyType: Serialize;
+
+    /// Send a request to a particular Cloudflare API endpoint, get the response as bytes.
+    async fn request_raw_bytes<ResultType, QueryType, BodyType>(
```
sssilver: What's the point of passing ResultType if you're gonna return Bytes?

sssilver: I don't agree with this notion that "any endpoint can be requested to return its actual return type, or to return a bunch of bytes". This feels like sloppy design to me. The reality of our API is that an endpoint always knows whether it returns a bunch of bytes or something else. That fact must be reflected in our API client design, there's no point in giving the user a choice between two functions to invoke.

cloudflare/src/framework/async_api.rs
```diff
@@ -55,20 +68,16 @@ impl Client {
...
-}
 
-#[async_trait]
-impl ApiClient for Client {
-    async fn request<ResultType, QueryType, BodyType>(
+    fn make_request<ResultType, QueryType, BodyType>(
         &self,
         endpoint: &(dyn Endpoint<ResultType, QueryType, BodyType> + Send + Sync),
-    ) -> ApiResponse<ResultType>
+    ) -> reqwest::RequestBuilder
```
sssilver: Making changes like this should be discussed before they become a PR. This changes the API of the library in a significant way.

cloudflare/src/framework/apiclient.rs
```diff
@@ -5,9 +5,9 @@ use crate::framework::{
 };
 use serde::Serialize;
 
-/// Synchronously sends requests to the Cloudflare API.
+/// Blocks and sends requests to the Cloudflare API.
```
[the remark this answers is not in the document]
adamchalmers: I was inspired by Reqwest which has named its APIs `reqwest::` and `reqwest::blocking::`.

I was thinking about why they did that. I like that the "default" API client is async, and to use the legacy synchronous API, you should have to type the word "blocking" because it makes you aware of the performance cost.

cloudflare/src/framework/async_api.rs
adamchalmers: We're not passing ResultType, we're acknowledging that this function can take an endpoint with any ResultType. After all, every ResultType can be expressed as bytes.

adamchalmers: I agree that some endpoints should never be requested as JSON. But I think it's reasonable for JSON endpoints to be requested as bytes. Some users may use cloudflare-rs for the strong type checking of endpoints and parameters, but want byte output. Or maybe they want to apply some function between receiving the response and deserializing the JSON (e.g. logging, filtering).

cloudflare/src/framework/async_api.rs
adamchalmers: I agree! I was just hacking away at this to see if it was possible, and what the implications of a design like this would be. Then I figured, well, if I have the code and it compiles, may as well have the discussion on the PR instead of just throwing the code away. I'm fine if we don't want to merge this :)

adamchalmers:
Closed in favor of https://github.com/cloudflare/cloudflare-rs/pull/87

PR #85, opened 2020-01-09
Trait for async API clients

adamchalmers:
This PR

- Adds an async API trait
- Makes the async API client impl it
- Makes the mock API client impl it
- Updates the e2e tests, to demonstrate that functions can take the async API trait as a parameter (instead of the async API client).

Note this adds a slight bit of overhead for using the async client, because of `dyn`, which has to box the concrete type and put it on the heap (I think). But that's going to be dominated by the cost of the network calls and possibly the serializing/deserializing JSON

### DIFF
--- a/cloudflare/src/framework/async_api.rs
+++ b/cloudflare/src/framework/async_api.rs
@@ -7,8 +7,21 @@ use crate::framework::{
     response::{ApiResponse, ApiResult},
     Environment, HttpApiClientConfig,
 };
+use async_trait::async_trait;
 use reqwest;
 use serde::Serialize;
+
+#[async_trait]
+pub trait ApiClient {
+    async fn request<ResultType, QueryType, BodyType>(
+        &self,
+        endpoint: &(dyn Endpoint<ResultType, QueryType, BodyType> + Send + Sync),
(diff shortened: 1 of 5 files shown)


cloudflare/src/framework/async_api.rs
```diff
@@ -42,12 +55,13 @@ impl Client {
             http_client,
         })
     }
+}
 
-    // Currently, futures/async fns aren't supported in traits, and won't be for a while.
-    // So unlike ApiClient/HttpApiClient, there's no trait with the request method.
-    pub async fn request<ResultType, QueryType, BodyType>(
+#[async_trait]
```
EverlastingBugstopper: I wonder if the language will ever support this directly - I wasn't able to find anything in the rust-lang issue tracker.

adamchalmers: The [async book](https://rust-lang.github.io/async-book/07_workarounds/06_async_in_traits.html) mentions it, and [this article explains why it's hard](https://smallcultfollowing.com/babysteps/blog/2019/10/26/async-fn-in-traits-are-hard/)

PR #87, opened 2020-01-12
Fix #26: Binary responses with stronger typing

### LINKED ISSUE #26: HttpApiClient should be able to handle non-JSON request and response bodies

adamchalmers:
This PR is based off [PR 84](https://github.com/cloudflare/cloudflare-rs/pull/84) but adds a new commit. Now you have to declare an endpoint as having binary or JSON responses. This is trying to address @sssilver's concern that endpoints should be defined as either binary or JSON, and only be used with one method, either `request` or `request_binary`. 

I added a `Binary` type which can never be instantiated and is not serializable. If you want an endpoint to return binary, use `ResultType = Binary`, so the compiler knows there's no JSON. You can see in the doctests I added that you can't use `ResultType = Binary` endpoints with `request`, and you can't use JSON endpoints with `request_binary`. So now users can't accidentally get a binary/JSON endpoint as JSON/binary :)

### DIFF
--- a/cloudflare/src/framework/apiclient.rs
+++ b/cloudflare/src/framework/apiclient.rs
@@ -1,6 +1,6 @@
 //! This module contains the synchronous (blocking) API client.
 use crate::framework::{
-    endpoint::Endpoint,
+    endpoint::{Binary, Endpoint},
     response::{ApiResponse, ApiResult},
 };
 use serde::Serialize;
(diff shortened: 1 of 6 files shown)


cloudflare/src/framework/apiclient.rs
```diff
@@ -16,4 +16,14 @@ pub trait ApiClient {
         ResultType: ApiResult,
         QueryType: Serialize,
         BodyType: Serialize;
+
+    /// Block and send a request to the Cloudflare API, get the response as bytes.
+    ///
+    fn request_binary<QueryType, BodyType>(
```
ashleymichal: for me, this puts the decision of whether a request should expect binary or not in the wrong place; as a user i think i should still be simply using `request`, and it should be up to the endpoint developers to specify what should come back, but this interface suggests that i might be able to call either request or request binary on any endpoint and that isn't so. an endpoint should be considered one or the other, but the method request might be better off taking an enum...

PR #91, opened 2020-01-23
Secrets endpoints

victoriabernard92:
Addition of the future Workers secret API endpoint

### DIFF
--- a/cloudflare/src/endpoints/workers/create_secret.rs
+++ b/cloudflare/src/endpoints/workers/create_secret.rs
@@ -1,0 +1,37 @@
+use super::WorkersSecret;
+
+use crate::framework::endpoint::{Endpoint, Method};
+
+/// Create Secret
+/// https://api.cloudflare.com/#worker-create-secret
+pub struct CreateSecret<'a> {
+    /// account id of owner of the script
+    pub account_identifier: &'a str,
+    /// the name of the script to attach the secret to
+    pub script_name: &'a str,
+    /// secert's contents
+    pub params: CreateSecretParams,
+}
+
+impl<'a> Endpoint<WorkersSecret, (), CreateSecretParams> for CreateSecret<'a> {
+    fn method(&self) -> Method {
+        Method::Put
+    }
+    fn path(&self) -> String {
+        format!(
+            "accounts/{}/workers/scripts/{}/secrets",
+            self.account_identifier, self.script_name
+        )
+    }
+    fn body(&self) -> Option<CreateSecretParams> {
+        Some(self.params.clone())
+    }
+}
+
+#[derive(Serialize, Clone, Debug)]
+pub struct CreateSecretParams {
+    /// the variable name of the secret that will be bound to the script
+    pub name: String,
+    /// the string value of the secret
+    pub value: String,
+}
--- a/cloudflare/src/endpoints/workers/create_secret.rs
+++ b/cloudflare/src/endpoints/workers/create_secret.rs
@@ -33,5 +33,7 @@ pub struct CreateSecretParams {
     /// the variable name of the secret that will be bound to the script
     pub name: String,
     /// the string value of the secret
-    pub value: String,
+    pub text: String,
+    // type of binding (e.g.secret_text)
+    pub r#type: String,
 }
--- a/cloudflare/src/endpoints/workers/create_secret.rs
+++ b/cloudflare/src/endpoints/workers/create_secret.rs
@@ -35,5 +35,6 @@ pub struct CreateSecretParams {
     /// the string value of the secret
     pub text: String,
     // type of binding (e.g.secret_text)
-    pub r#type: String,
+    #[serde(rename = "type")]
+    pub secret_type: String,
 }
--- a/cloudflare/src/endpoints/workers/mod.rs
+++ b/cloudflare/src/endpoints/workers/mod.rs
@@ -23,13 +23,13 @@ pub use list_secrets::ListSecrets;
 /// https://api.cloudflare.com/#worker-routes-properties
 #[derive(Deserialize, Serialize, Debug, Clone, PartialEq, Eq)]
 pub struct WorkersRoute {
-  /// Namespace identifier tag.
-  pub id: String,
-  /// The basic pattern that should map to the script
-  pub pattern: String,
-  /// Name of the script to apply when the route is matched.
-  /// The route is skipped when this is blank/missing.
-  pub script: Option<String>,
+    /// Namespace identifier tag.
+    pub id: String,
+    /// The basic pattern that should map to the script
+    pub pattern: String,
+    /// Name of the script to apply when the route is matched.
+    /// The route is skipped when this is blank/missing.
+    pub script: Option<String>,
 }
 
 impl ApiResult for WorkersRoute {}
@@ -40,8 +40,8 @@ impl ApiResult for Vec<WorkersRoute> {}
 /// but it feels wrong.
 #[derive(Deserialize, Serialize, Debug, Clone, PartialEq, Eq)]
 pub struct WorkersRouteIdOnly {
-  /// Namespace identifier tag.
-  pub id: String,
+    /// Namespace identifier tag.
+    pub id: String,
 }
 
 impl ApiResult for WorkersRouteIdOnly {}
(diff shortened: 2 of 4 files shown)


cloudflare/src/endpoints/workers/mod.rs
```diff
@@ -3,7 +3,9 @@ use crate::framework::response::ApiResult;
 use serde::Deserialize;
 
 mod create_route;
+pub mod create_secret;
 mod delete_route;
+pub mod delete_secret;
```
ashleymichal: i know  you said you had given it  a try earlier, but i'd like to do the same `pub use` as we did with routes endpoints. the  reason being that the way this will look in client code is:

``` rust
// we prefer this
use cloudflare::endpoints::workers::{CreateRoute, CreateRouteParams}
// over this
use cloudflare::endpoints::workers::create_secret::{CreateSecret, CreateSecretParams}
```

cloudflare/src/endpoints/workers/mod.rs
```diff
@@ -37,3 +39,16 @@ pub struct WorkersRouteIdOnly {
...
 impl ApiResult for WorkersRouteIdOnly {}
+
+/// Secrets attach to a single script to be readable in only the script
+/// https://api.cloudflare.com/#worker-secrets-properties
+#[derive(Deserialize, Serialize, Debug, Clone, PartialEq, Eq)]
+pub struct WorkersSecret {
+  /// TODO these fields depend on the API and may be wrong
+  pub name: String,
+  pub modified_on: String,
+  pub script_id: String,
```
ashleymichal: we won't be returning script id, but we will return created on.

cloudflare/src/endpoints/workers/create_secret.rs
```diff
@@ -0,0 +1,37 @@
...
+
+use crate::framework::endpoint::{Endpoint, Method};
+
+/// Create Secret
+/// https://api.cloudflare.com/#worker-create-secret
+pub struct CreateSecret<'a> {
+    /// account id of owner of the script
+    pub account_identifier: &'a str,
+    /// the name of the script to attach the secret to
+    pub script_name: &'a str,
```
sssilver: I'm starting to think that these should be owned strings

cloudflare/src/endpoints/workers/mod.rs
```diff
@@ -37,3 +39,16 @@ pub struct WorkersRouteIdOnly {
...
 
 impl ApiResult for WorkersRouteIdOnly {}
+
+/// Secrets attach to a single script to be readable in only the script
+/// https://api.cloudflare.com/#worker-secrets-properties
+#[derive(Deserialize, Serialize, Debug, Clone, PartialEq, Eq)]
+pub struct WorkersSecret {
+  /// TODO these fields depend on the API and may be wrong
+  pub name: String,
+  pub modified_on: String,
```
sssilver: This should probably be a datetime of some sort

cloudflare/src/endpoints/workers/create_secret.rs
```diff
@@ -0,0 +1,37 @@
...
+        Some(self.params.clone())
+    }
+}
+
+#[derive(Serialize, Clone, Debug)]
+pub struct CreateSecretParams {
+    /// the variable name of the secret that will be bound to the script
+    pub name: String,
+    /// the string value of the secret
+    pub value: String,
```
sssilver: Are secret values strings or bytes? It feels like the latter would be more proper, no?

cloudflare/src/endpoints/workers/mod.rs
victoriabernard92: I try to do that but it kept complaining it was private or not defined. Could you point me to how you used it in wrangler routes code?

cloudflare/src/endpoints/workers/mod.rs
```diff
@@ -44,10 +48,11 @@ impl ApiResult for WorkersRouteIdOnly {}
 /// https://api.cloudflare.com/#worker-secrets-properties
 #[derive(Deserialize, Serialize, Debug, Clone, PartialEq, Eq)]
 pub struct WorkersSecret {
-  /// TODO these fields depend on the API and may be wrong
+  /// TODO: these fields depend on the API and may be wrong since unable to test
   pub name: String,
-  pub modified_on: String,
-  pub script_id: String,
+  pub r#type: String,
```
sssilver: what's this syntax? 😮

victoriabernard92: type is a reserved name so my text editor recommended this. 
image

cloudflare/src/endpoints/workers/create_secret.rs
victoriabernard92: What do you mean by these? @ashleymichal how should we apply this suggestion?

sssilver: Basically turn `&'a str` into a `String`

ashleymichal: we don't do this anywhere else :-\

cloudflare/src/endpoints/workers/create_secret.rs
ashleymichal: yah bytes is probably better

cloudflare/src/endpoints/workers/create_secret.rs
```diff
@@ -0,0 +1,39 @@
...
+}
+
+#[derive(Serialize, Clone, Debug)]
+pub struct CreateSecretParams {
+    /// the variable name of the secret that will be bound to the script
+    pub name: String,
+    /// the string value of the secret
+    pub text: String,
+    // type of binding (e.g.secret_text)
+    pub r#type: String,
```
sssilver: I would suggest just calling this `secret_type`, though I'll defer to the other guys as I don't feel strongly about it

ashleymichal: elsewhere in the codebase we have used `serde_rename`, and i'm down with doing that.

cloudflare/src/endpoints/workers/create_secret.rs
ashleymichal: except you can't precisely send a byte string in a json body...

cloudflare/src/endpoints/workers/create_secret.rs
ashleymichal: my bad, this is `serde(rename = "foo")`

cloudflare/src/endpoints/workers/mod.rs
```diff
@@ -1,27 +1,35 @@
...
 #[derive(Deserialize, Serialize, Debug, Clone, PartialEq, Eq)]
 pub struct WorkersRoute {
-    /// Namespace identifier tag.
-    pub id: String,
-    /// The basic pattern that should map to the script
-    pub pattern: String,
-    /// Name of the script to apply when the route is matched.
-    /// The route is skipped when this is blank/missing.
-    pub script: Option<String>,
+  /// Namespace identifier tag.
```
ashleymichal: looks like your linter is messing with tabs. tab size should be 4 spaces. seems to be only this file? idk, make sure you check all files.

cloudflare/src/endpoints/workers/create_secret.rs
```diff
@@ -0,0 +1,40 @@
...
+use crate::framework::endpoint::{Endpoint, Method};
+
+/// Create Secret
+/// https://api.cloudflare.com/#worker-create-secret
+pub struct CreateSecret<'a> {
+    /// account id of owner of the script
+    pub account_identifier: &'a str,
+    /// the name of the script to attach the secret to
+    pub script_name: &'a str,
+    /// secert's contents
```
adamchalmers: s/secert/secret

Also all our comments elsewhere start with a capital, could you please change yours to be consistent? :)

cloudflare/src/endpoints/workers/create_secret.rs
sssilver: You can serialize them as a string, however within the Rust type system it should still be bytes

sssilver: That would force you to think about how that "string" should be deserialized. There isn't a single way to serialize bytes into a string.

cloudflare/src/endpoints/workers/mod.rs
sssilver: I'm not opposed to using core language features to solve problems

victoriabernard92: I took your suggestion

cloudflare/src/endpoints/workers/create_secret.rs
victoriabernard92: This was giving me a hard time. Since it's not a blocker I filed an issue to prioritize later https://github.com/cloudflare/cloudflare-rs/issues/98

cloudflare/src/endpoints/workers/create_secret.rs
ashleymichal: @sssilver i actually started turning all of these `&'a str`'s into `&'static str`'s. saves the requirement for lifetimes on all your structs and their methods. will illustrate in another pr.

Issue #94, opened 2020-01-28
Creating overview implemented endpoints - Unknown endpoint user/tokens/verify


SvenJanssens:
Hi,

I have created a version of README mentioning which API endpoints are implemented (and which not).
However I don't find the "Validate User Token" one in the API documentation at https://api.cloudflare.com . Is this a new one?

Have a nice day!

Regards,

Sven

sssilver:
Hi @SvenJanssens -- thanks for the lovely contribution, this is very cool!

All you need to do is just create a PR against the origin repo, and we would review it and get it merged.

PR #95, opened 2020-01-28
Add Implemented Endpoints to README

SvenJanssens:
Add Implemented Endpoints to README - Issue #94

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,3 +7,655 @@
 ⚠️ This library is a Work in Progess! ⚠️
 
 This library provides convenience functions that wrap the Cloudflare API.
+
+## Implemented Endpoints
+[User](https://api.cloudflare.com/#user-properties) - The currently logged in/authenticated User
+- [X] [User Details](https://api.cloudflare.com/#user-user-details)
+- [ ] [Edit Details](https://api.cloudflare.com/#user-edit-user) - Edit part of your user details
+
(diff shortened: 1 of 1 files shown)


SvenJanssens:
Hi,
I have created the PR. Not sure who are usually the reviewers, so I have assigned @sssilver and @nickvollmar (as I saw the same reviewers were also assigned to previous PRs).
Thanks a lot in advance and have a nice day!
Regards,
Sven

ashleymichal:
wow this is exhaustive, well done! while i appreciate the utility, i am not confident this will be well maintained, or as useful as it could be; i think it would be easier/less noisy to either list what we DO implement, or somehow script a way to show what we do/don't implement. that's my two cents anyway.

SvenJanssens:
I actually made the complete list to make it easier to maintain. As for now we can just put the X between the brackets to mark something as implemented. While if it just lists the endpoints implemented, then more effort is needed, causing more friction and so it becomes less likely for it to get maintained. A script would be even better, but I am not at home with writing one.
If it would create too much noise on the readme, then I don't mind just keeping it as an overview for myself on my branch. Should I close the pull request?

SvenJanssens:
Hmm, looks like I messed up the pull request by updating the branch with the latest changes. So I will close it. Sorry for the troubles.

ashleymichal:
i don't think you need to close it, i think others  should chime in as well. i'm only one voice and  the utility is helpful for sure.

PR #105, opened 2020-02-09
Fixes #26. Allow endpoints to specify a serialization function.

### LINKED ISSUE #26: HttpApiClient should be able to handle non-JSON request and response bodies

ispivey:
This change allows endpoints to specify their own body serialization function, while preserving the current default behavior with a default function in the Endpoint trait, addressing #26 . 

A subsequent PR will demonstrate how this can be used to implement the "Create or update Worker" API endpoint.

More detail in my comment here: https://github.com/cloudflare/cloudflare-rs/issues/26#issuecomment-583869870

### DIFF
--- a/cloudflare/src/framework/endpoint.rs
+++ b/cloudflare/src/framework/endpoint.rs
@@ -25,6 +25,12 @@ where
     fn body(&self) -> Option<BodyType> {
         None
     }
+    fn serialized_body(&self) -> Option<String> {
+        match self.body() {
+            Some(body) => Some(serde_json::to_string(&body).unwrap()),
+            None => None
+        }
+    }
     fn url(&self, environment: &Environment) -> Url {
         Url::from(environment).join(&self.path()).unwrap()
     }
--- a/cloudflare/src/framework/mod.rs
+++ b/cloudflare/src/framework/mod.rs
@@ -116,8 +116,8 @@ impl<'a> ApiClient for HttpApiClient {
             )
             .query(&endpoint.query());
 
-        if let Some(body) = endpoint.body() {
-            request = request.body(serde_json::to_string(&body).unwrap());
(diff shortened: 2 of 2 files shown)


ispivey:
Hrm, thinking about it a little more - I'm not sure if this takes us in the right direction, if the next step is being able to specify a multi-part form body. 

To get that, the `Endpoint` needs to be able to tell `framework::ApiClient` that it should construct a `reqwest::multipart::Form` and set the request body with `request.multipart(form)` rather than `request.body(...)`. Additionally, it needs to specify the parts of the form in excruciating detail (as @ashleymichal pointed out in #26 , seemingly years ago when I was a naive child and had not wrestled with this for a while).

A couple ideas:
1. Make Endpoints capable of specifying exactly how the body is constructed. They could have a function called by `ApiClient::request` that takes a `RequestBuilder`, adds the body in the desired manner, and returns a `RequestBuilder` that `ApiClient::request` can further modify and send. This has the downside of coupling `Endpoint` with `reqwest`, and I assume they're purposefully decoupled today. However, it would allow an endpoint like `CreateWorker` to make a multi-part form body explicitly using `reqwest`.
2. Add another `enum` to specify body type (json, string, form, multi-part form) that can be specified as part of an `Endpoint` implementation. This is analogous to the `endpoint::Method` enum, and thus seems more in line with the library's current approach. However, it would require creating another layer of indirection for the `Endpoint` to specify how its form should be built, and for `ApiClient::request` to read and act on that specification.

Gonnna go take a break from this and talk about it on Monday 😬

sssilver:
Thanks for the PR! This subject has been coming up a few times, and at this point I think it'd be good for us to have a technical discussion where we all converge on a single solution.

kennetheversole:
Hello, 

PR #112, opened 2020-03-04
Workers tailing endpoints

gabbifish:
This PR adds API endpoints for upcoming Workers trace logging functionality.

I would still like to add formal tests before merging this into cloudflare-rs and doing a release.

### DIFF
--- a/cloudflare/src/endpoints/workers/create_tail_heartbeat.rs
+++ b/cloudflare/src/endpoints/workers/create_tail_heartbeat.rs
@@ -1,0 +1,26 @@
+use super::WorkersTail;
+
+use crate::framework::endpoint::{Endpoint, Method};
+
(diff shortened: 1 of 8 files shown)


cloudflare/src/endpoints/workers/create_tail_heartbeat.rs
```diff
@@ -0,0 +1,26 @@
+use super::WorkersTail;
+
+use crate::framework::endpoint::{Endpoint, Method};
+
+/// Send Tail Heartbeat
+/// https://api.cloudflare.com/#worker-tail-heartbeat
+pub struct CreateTailHeartbeat<'a> {
```
ashleymichal: i think i'd actually like to call this "SendTailHeartbeat"

PR #115, opened 2020-03-28
add default for account struct

shalzz:
When querying the list zones api "GET zones" https://api.cloudflare.com/#zone-list-zones
The account details returned only contain the `id` and the `name` fields therefore create placeholder defaults for the remaining fields in the `Account` struct.

### DIFF
--- a/cloudflare/src/endpoints/account/mod.rs
+++ b/cloudflare/src/endpoints/account/mod.rs
@@ -28,5 +28,16 @@ pub struct Settings {
     enforce_twofactor: bool,
 }
 
+/// Cloudflare Accounts Details
+/// An Account is the root object which owns other resources such as zones, load balancers and billing details.
+/// https://api.cloudflare.com/#accounts-properties
+#[derive(Deserialize, Serialize, Debug, Clone, PartialEq, Eq)]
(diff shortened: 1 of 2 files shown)


cloudflare/src/endpoints/account/mod.rs
```diff
@@ -15,18 +15,24 @@ pub struct Account {
     /// Account name
     pub name: String,
     /// Account Settings
+    #[serde(default)]
     pub settings: Settings,
     /// describes when the account was created
+    #[serde(default = "default_time")]
```
sssilver, line 21: What's the value of having the time when account was created default to now()?

adamchalmers (changes requested):
Rather than add defaults for the Account struct, I'd rather we change the [Zone::account](https://docs.rs/cloudflare/0.6.6/cloudflare/endpoints/zone/struct.Zone.html#structfield.account) field so that it's no longer of type Account, but rather some new type with only `id` and `name`, probably named `AccountDetails` or `AccountIdName`.

adamchalmers (approved):

nataliescottdavidson (approved):

ashleymichal (approved):

PR #116, opened 2020-03-28
Handle getting an empty frequency field for Free Plan

shalzz:
The `frequency` field is empty when on the free plan.

### DIFF
--- a/cloudflare/src/endpoints/plan.rs
+++ b/cloudflare/src/endpoints/plan.rs
@@ -1,5 +1,6 @@
 #[derive(Deserialize, Debug)]
 #[serde(rename_all = "lowercase")]
+/// Free plans won't have a Frequency, so most responses should accept Option instead.
 pub enum Frequency {
     Weekly,
     Monthly,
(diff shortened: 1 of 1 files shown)


cloudflare/src/endpoints/plan.rs
```diff
@@ -5,6 +5,9 @@ pub enum Frequency {
     Monthly,
     Quarterly,
     Yearly,
+    /// We get an empty variant for free plans
+    #[serde(other)]
+    Missing,
```
adamchalmers, line 10: Rather than add a new case for missing, I'd rather structs just use Option<Frequency> instead, and update the docstring for this enum to say "Free plans won't have a Frequency, so most responses should accept Option<Frequency> instead."

adamchalmers (changes requested):

cloudflare/src/endpoints/plan.rs
shalzz: Done

adamchalmers (approved):

ashleymichal (approved):

PR #119, opened 2020-05-07
correct Workers DELETE /tail endpoint return type

ashleymichal:
the api returns an empty result body. this has been tested locally with  wrangler tail.


ashleymichal:
@ashleygwilliams probably!

PR #121, opened 2020-07-31
adds staging to environment enum

EverlastingBugstopper:
`Environment::Staging` instead of `Environment::Custom(Url::parse("https://api.staging.cloudflare.com/v4").unwrap()))` bc common use case

### DIFF
--- a/cloudflare/src/framework/mod.rs
+++ b/cloudflare/src/framework/mod.rs
@@ -37,6 +37,7 @@ pub enum SearchMatch {
 #[derive(Debug)]
 pub enum Environment {
     Production,
+    Staging,
     Custom(url::Url),
 }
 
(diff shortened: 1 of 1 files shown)


cloudflare/src/framework/mod.rs
```diff
@@ -45,6 +46,9 @@ impl<'a> From<&'a Environment> for url::Url {
         match environment {
             Environment::Production => {
                 url::Url::parse("https://api.cloudflare.com/client/v4/").unwrap()
+            },
+            Environment::Staging => {
+                url::Url::parse("https://api.staging.cloudflare.com/client/v4/").unwrap()
```
sssilver, line 51: Instead of coding this into the crate source, why not use `Environment::Custom`?

EverlastingBugstopper: for the same reason i wouldn't use `Environment::Custom` for production - it's a common use case for folks using this library and having the shorthand is convenient

PR #127, opened 2020-11-25
WIP - support metadata (depth 1)

ashleymichal:


cloudflare/src/endpoints/workerskv/write_bulk.rs
```diff
@@ -34,4 +36,6 @@ pub struct KeyValuePair {
     pub expiration: Option<i64>,
     pub expiration_ttl: Option<i64>,
     pub base64: Option<bool>,
+    // a temporary hack for basic metadata
+    pub metadata: Option<HashMap<String, String>>,
```
adamchalmers, line 40: Is there a meaningful distinction here between a hashmap with no values, and not having a hashmap?

If not, I think 
```rust
#[serde(default)]
pub metadata: HashMap<String, String>,
```
makes more sense, because otherwise you're doing two checks (is option.IsSome, then hashmap.IsEmpty) but this way you eliminate the optional check, and you can safely, idk, loop or map over the hashmap without worrying about the nil case.

edmundhung:
Hello! I am a wrangler user and realise the metadata support have to come from the rust library here. May I ask if there are any blockers/concerns on supporting it? I would be happy to give it a try if possible :)

cloudflare/src/endpoints/workerskv/write_bulk.rs
edmundhung: I have played with the API a bit today and realize there is actually no limitation on the shape of the metadata at all (e.g. a numeric array is also accepted). Would it be a bad idea to set it as a HashMap in this case?

PR #129, opened 2020-12-03
Make cloudflare crate usable in wasm32 targets

nmldiegues:
Using the cloudflare in a wasm32 target was failing to compile.
    The first reason being that the blocking client cannot ever work
    in wasm.
    The second reason being that the asynchronous client was using
    the async_trait that did not work fine under wasm_bindgen either.

    As a result, we now hide away those portions of the framework
    implementation when targeting wasm32. We do so by resorting to
    conditional compilation macros.

    To test this out we used the framework to call some endpoints in
    a Cloudflare Worker written in Rust (thus running in wasm) and it
    all worked fine after these changes.

    This thus enables Workers' developers in Rust to use this library
    at will and contribute further to it.

### DIFF
--- a/cloudflare/Cargo.toml
+++ b/cloudflare/Cargo.toml
@@ -13,6 +13,7 @@ anyhow = "1.0"
 async-trait = "0.1"
 base64 = "0.13"
 chrono = { version = "0.4", features = ["serde"] }
+cfg-if = "0.1"
 http = "0.2"
 percent-encoding = "1.0"
 reqwest = { version = "0.10", features = ["json", "blocking"] }
--- a/cloudflare/src/framework/async_api.rs
+++ b/cloudflare/src/framework/async_api.rs
@@ -10,6 +10,7 @@ use crate::framework::{
 use async_trait::async_trait;
 use reqwest;
 use serde::Serialize;
+use cfg_if::cfg_if;
 
 #[async_trait]
 pub trait ApiClient {
@@ -45,10 +46,18 @@ impl Client {
         config: HttpApiClientConfig,
         environment: Environment,
     ) -> anyhow::Result<Client> {
-        let http_client = reqwest::Client::builder()
-            .timeout(config.http_timeout)
-            .default_headers(config.default_headers)
-            .build()?;
+        #[allow(unused_mut)]
+        let mut builder = reqwest::Client::builder()
+            .default_headers(config.default_headers);
+
+        cfg_if! {
+            // There are no timeouts in wasm.
+            if #[cfg(not(target_arch = "wasm32"))] {
+                builder = builder.timeout(config.http_timeout);
+            }
+        }
+
+        let http_client = builder.build()?;
 
         Ok(Client {
             environment,
@@ -56,11 +65,8 @@ impl Client {
             http_client,
         })
     }
-}
 
-#[async_trait]
-impl ApiClient for Client {
-    async fn request<ResultType, QueryType, BodyType>(
+    pub async fn request_handle<ResultType, QueryType, BodyType>(
         &self,
         endpoint: &(dyn Endpoint<ResultType, QueryType, BodyType> + Send + Sync),
     ) -> ApiResponse<ResultType>
@@ -86,6 +92,23 @@ impl ApiClient for Client {
         request = request.auth(&self.credentials);
         let response = request.send().await?;
         map_api_response(response).await
+    }
+}
+
+// The async_trait does not work nicely in wasm.
+#[cfg(not(target_arch = "wasm32"))]
+#[async_trait]
+impl ApiClient for Client {
+    async fn request<ResultType, QueryType, BodyType>(
+        &self,
+        endpoint: &(dyn Endpoint<ResultType, QueryType, BodyType> + Send + Sync),
+    ) -> ApiResponse<ResultType>
+        where
+            ResultType: ApiResult,
+            QueryType: Serialize,
+            BodyType: Serialize,
+    {
+        self.request_handle(endpoint).await
     }
 }
 
--- a/cloudflare/src/framework/async_api.rs
+++ b/cloudflare/src/framework/async_api.rs
@@ -51,7 +51,7 @@ impl Client {
             .default_headers(config.default_headers);
 
         cfg_if! {
-            // There are no timeouts in wasm.
+            // There are no timeouts in wasm. The property is documented as no-op in wasm32.
             if #[cfg(not(target_arch = "wasm32"))] {
                 builder = builder.timeout(config.http_timeout);
             }
@@ -95,7 +95,8 @@ impl Client {
     }
 }
 
-// The async_trait does not work nicely in wasm.
+// The async_trait does not work nicely in wasm. The mapping of Rust Futures to wasm bindgen
+// Promises does not seem to work when the async_trait macro is used: it causes compilation failures.
 #[cfg(not(target_arch = "wasm32"))]
 #[async_trait]
 impl ApiClient for Client {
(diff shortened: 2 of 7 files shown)


nmldiegues:
The bi-product of this PR is that I could get a Rust Worker, stick in this to Cargo.toml:
```
cloudflare = { path = "../cloudflare-rs/cloudflare", default-features = false }
```

...and it compiled.


If I let it have the default features, it fails to compile due to missing reqwest stuff related with blocking as well as mapping Futures into wasm bindgen.

cloudflare/Cargo.toml
```diff
@@ -8,6 +8,10 @@ keywords = ["cloudflare", "api", "client"]
 categories = ["api-bindings", "web-programming::http-client"]
 license = "BSD-3-Clause"
 
+[features]
+default = [ "reqwest-http" ]
+reqwest-http = []
```
nmldiegues, line 13: Where should I document the effect of this feature?

cloudflare/src/framework/apiclient.rs
```diff
@@ -1,9 +1,10 @@
-//! This module contains the synchronous (blocking) API client.
+//! This module contains the API for a client.
```
nmldiegues, line 1: Self-comment:  I've modified this file to solely contain the traits for the sync and async clients.  That way I moved the implementations to separate files so that I can cleanly elide them with a cfg macro over the reqwest-http feature.

cloudflare/src/framework/apiclient.rs
```diff
@@ -17,3 +18,15 @@ pub trait ApiClient {
         QueryType: Serialize,
         BodyType: Serialize;
 }
+
+#[async_trait]
+pub trait AsyncApiClient {
```
nmldiegues, line 23: Self-comment: this was the previous async_api::ApiClient , which I renamed here to disambiguate with the sync client

cloudflare/src/framework/auth.rs
```diff
@@ -24,12 +22,3 @@ impl Credentials {
 pub trait AuthClient {
     fn auth(self, credentials: &Credentials) -> Self;
 }
-
-impl AuthClient for RequestBuilder {
```
nmldiegues, line 28: Self-comment: moved this into the blocking_api.rs that contains the implementation for a blocking client.

cloudflare/src/framework/blocking_api.rs
```diff
@@ -0,0 +1,97 @@
+use crate::framework::{Environment, auth, endpoint, response};
```
nmldiegues, line 1: Self-comment: new file, albeit all the comments were just moved, none of the code is new.

cloudflare/src/framework/config.rs
```diff
@@ -0,0 +1,18 @@
+use std::time::Duration;
```
nmldiegues, line 1: Self-comment: new file, but all of the comments were moved, none of them are new.

cloudflare/src/framework/mod.rs
```diff
@@ -2,17 +2,20 @@
 This module controls how requests are sent to Cloudflare's API, and how responses are parsed from it.
  */
 pub mod apiclient;
+#[cfg(feature = "reqwest-http")]
```
nmldiegues, line 5: Self-comment: and this is what we wanted to achieve, we can now cleanly elide a few modules away and get the crate to compile under wasm.

Without the changes in this PR this would be a mess of macros all around due to the more promiscuous coupling of reqwest that was all around.
it's still not perfect, but good enough for now

nmldiegues (commented):
Self commented a few bits to help clarify the review process.

Note that this was a pure refactoring. There was no logic changed whatsover.

cloudflare/src/framework/apiclient.rs
adamchalmers: Let's not move symbols into new modules -- it's a breaking change and doesn't seem necessary. Same with all the other moves. Ideally this PR would not cause any breaking changes (i.e. would not move any symbols out of their current modules). Adding new symbols is fine though.

adamchalmers (changes requested):

cloudflare/src/framework/apiclient.rs
nmldiegues: I've reverted back the breaking changes, resulting in a smaller diff. So that's good news.

The downside is that some of the changes did make sense, i.e., things do seem a little out of place, but I guess we can live with that. Furthermore, this requires spreading around some random `#[cfg(feature = "reqwest-http")]` which is not great, but it's the cost of backwards compatibility.

chungthuang (approved):

nmldiegues:
This PR changed quite a bit (I could as well have abandoned and opened a new one...).

Modified title and message, which I think are comprehensive enough to explain the change in course for the PR.

Thanks for the re-review!

cloudflare/src/framework/async_api.rs
```diff
@@ -45,22 +46,27 @@ impl Client {
...
-        let http_client = reqwest::Client::builder()
-            .timeout(config.http_timeout)
-            .default_headers(config.default_headers)
-            .build()?;
+        #[allow(unused_mut)]
+        let mut builder = reqwest::Client::builder()
+            .default_headers(config.default_headers);
+
+        cfg_if! {
+            // There are no timeouts in wasm.
```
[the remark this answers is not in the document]
nmldiegues: Correct. I will add a note to the the doc of that property in the struct `HttpApiClientConfig`.

cloudflare/src/framework/blocking_api.rs
```diff
@@ -0,0 +1,72 @@
...
+        Ok(HttpApiClient {
+            environment,
+            credentials,
+            http_client,
+        })
+    }
+}
+
+// TODO: This should probably just implement request for the Reqwest client itself :)
+// TODO: It should also probably be called `ReqwestApiClient` rather than `HttpApiClient`.
```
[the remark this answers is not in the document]
nmldiegues: I wouldn't know, maybe @adamchalmers  can chime in.

Based on his feedback about the other stuff, I think we'd prefer to not break symbols unnecessarily that would cause users to have to adapt their code when updating cloudflare-rs version.
Still, my guess is only as good as yours (or even less).

But given that I'm only moving the implementation to a new file, as the whole PR is a non-breaking mini-refactor from the perspective of non-wasm targets, I did not want to meddle with what already existed here.

cloudflare/src/framework/async_api.rs
```diff
@@ -89,6 +95,23 @@ impl ApiClient for Client {
     }
 }
 
+// The async_trait does not work nicely in wasm.
```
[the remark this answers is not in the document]
nmldiegues: I've added a couple comments to expand on this. I've shown it briefly to @adamchalmers  and it seems like something would need fixing (in either async_trait or wasm_bindgen crates, or both). For the time being this is a good enough compromise.

cloudflare/src/framework/mod.rs
```diff
@@ -4,14 +4,15 @@ This module controls how requests are sent to Cloudflare's API, and how response
 pub mod apiclient;
 pub mod async_api;
 pub mod auth;
+#[cfg(not(target_arch = "wasm32"))]  // There is no blocking implementation for wasm.
+pub mod blocking_api;
```
adamchalmers, line 8: Can you please add a `pub use blocking_api::*` so that this doesn't break backwards compatibility?

nmldiegues: I don't think this breaks compatibility since that file only contains "impl" blocks, i.e., no public symbols.

adamchalmers (approved):

PR #130, opened 2020-12-03
New endpoint for creating Named Argo Tunnels

nmldiegues:


cloudflare/src/endpoints/argo_tunnel/create_tunnel.rs
```diff
@@ -0,0 +1,36 @@
...
+        format!("accounts/{}/tunnels", self.account_identifier)
+    }
+    fn body(&self) -> Option<CreateTunnelParams<'a>> {
+        Some(self.params.clone())
+    }
+}
+
+/// Params for creating a Named Argo Tunnel
+#[derive(Serialize, Clone, Debug)]
+pub struct CreateTunnelParams<'a> {
```
adamchalmers, line 29: I suggest naming this Params, so that users can write `create_tunnel::Params`. Like the example of `create_pool::Params`.

cloudflare/src/endpoints/argo_tunnel/mod.rs
```diff
@@ -0,0 +1,29 @@
...
+
+use crate::framework::response::ApiResult;
+
+pub mod create_tunnel;
+
+/// A Named Argo Tunnel
+/// This is an Argo Tunnel that has been created. It can be used for routing and subsequent running.
+/// https://api.cloudflare.com/#argo-tunnel-properties
+#[derive(Deserialize, Serialize, Debug, Clone, PartialEq, Eq)]
+pub struct Tunnel {
```
adamchalmers, line 12: Suggestion/nit: Move these into a `data_structures.rs` file in this module, and then have this mod.rs `pub use datastructures::*` so all the datatypes are in the main `argo_tunnel` namespace.

adamchalmers (changes requested):

adamchalmers (approved):

adamchalmers:
LGTM

cloudflare/src/framework/json_utils.rs
```diff
@@ -0,0 +1,34 @@
...
+where
+    D: serde::Deserializer<'de>,
+{
+    struct Base64Visitor;
+
+    impl<'de> de::Visitor<'de> for Base64Visitor {
+        type Value = Vec<u8>;
+
+        fn expecting(&self, formatter: &mut ::std::fmt::Formatter) -> ::std::fmt::Result {
+            write!(formatter, "base64 ASCII text")
```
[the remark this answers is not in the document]
nmldiegues: This method is not the "interesting" part. It is used only to create an error message stating what is expected during deserialization, such that if the method below fails or is never called by the "AST traversal", then this one is invoked.

See https://docs.serde.rs/serde/de/trait.Visitor.html#tymethod.expecting

And yes, "base64 ASCII text" is the error message that would appear under an error that says:
```
This Visitor expects to receive base64 ASCII text
```
in the context of a serde error.

chungthuang (approved):

PR #136, opened 2021-02-07
(fix master) Use Result from std instead of serde's undocumented now private re-export

vvilhonen:
Also enables github action test runs on pull requests


nmldiegues:
That's odd. I just did a clean cargo build over a new clone of master and it built locally.

Still, this PR makes sense, thanks for doing it.

nmldiegues (approved):

sssilver (approved):

PR #140, opened 2021-04-14
Update tail endpoint

caass:
There's now a V2 implementation of tail that doesn't involve an external URL, so that param should be optional. The API respects an empty body, an empty JSON object body, and no body, so whatever I managed to do here should be fine. If I understand correctly, the way I've written this will omit the body entirely.

### DIFF
--- a/cloudflare/src/endpoints/workers/create_tail.rs
+++ b/cloudflare/src/endpoints/workers/create_tail.rs
@@ -10,7 +10,11 @@ pub struct CreateTail<'a> {
     pub account_identifier: &'a str,
     /// The name of the script to tail
     pub script_name: &'a str,
-    /// URL to which to send events
+    /// V1 of tailing involved creating a separate URL,
+    /// which is still possible.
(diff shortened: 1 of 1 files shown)


cloudflare/src/endpoints/workers/create_tail.rs
```diff
@@ -25,12 +29,16 @@ impl<'a> Endpoint<WorkersTail, (), CreateTailParams> for CreateTail<'a> {
...
+        if self.params.url.is_some() {
+            Some(self.params.clone())
+        } else {
+            None
+        }
     }
 }
 
 #[derive(Serialize, Clone, Debug)]
 pub struct CreateTailParams {
```
adamchalmers, line 41: I suggest this object should now derive Default, because its only field includes Default.

adamchalmers (changes requested):

adamchalmers (approved):

sssilver (approved):

PR #144, opened 2021-05-10
fix: handle possible missing betas field from user details response

nilslice:
In response to a recent upstream change within the public API, the Wrangler client is seeing some errors, e.g. https://github.com/cloudflare/wrangler/issues/1914

This PR provides a default to the `betas` field which now may likely be empty (and omitted from the API response) in many more cases. The attribute added here is done so that we are not forcing a change on users of this crate to otherwise handle an `Option<Vec<String>>`, but rather encounter an empty vec if `betas` is not present in the json response.

### DIFF
--- a/cloudflare/Cargo.toml
+++ b/cloudflare/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "cloudflare"
-version = "0.8.1"
+version = "0.8.2"
 authors = ["Areg Harutyunyan <areg@cloudflare.com>"]
 edition = "2018"
 description = "Rust library for the Cloudflare v4 API"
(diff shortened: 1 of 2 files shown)


nilslice:
PS - this does not include any version change in Cargo.toml. Happy to update it, just want to determine the version number we want to release.

nilslice:
Just for posterity:
> This is technically a breaking change for any user who instantiated the struct by specifying all fields. Users who used the ..default() method to instantiate won't see any change. I guess technically this requires a major version bump, but that seems way too scary.

Adam and I discussed offline, we're all good - no breaking change 😄 . The only change here is the addition of the attribute, not the field itself.

PR #150, opened 2021-06-18
 Rewrite core framework to support new "request-response builder" design

ObsidianMinor:
As discussed internally, this is my desired rewrite of the cloudflare-rs library. It was made with ease of use and simplicity in mind and should hopefully open up the library to some endpoints not currently supported today. It primarily does this by inverting control of serialization and allowing endpoints to call into a generic request serialization and response deserialization system that can be expanded upon in the future should other endpoints that aren't supported by it be created.

Because it rewrites most of the core library, all the endpoints needed to be rewritten as well, so I reorganized them to cut down on duplicate naming information and combine multiple relevant endpoints into their own self contained module (ex: `cloudflare::endpoints::load_balancing::create_pool::CreatePool` becomes `cloudflare::api::load_balancer::pools::Create`). On top of this, I made all endpoints, json bodies, queries, and json responses have a defined standard set of included traits for consistency:

* Endpoints (with URL path parameters) implement Debug, Clone, and PartialEq
* Responses, bodies, and queries (including endpoints where their endpoint definition is itself the body or query) implement Deserialize, Serialize, Debug, Clone, and PartialEq and never have lifetime parameters.
* Default is implemented whenever possible.

However, with this new design, some features are lost. Because clients don't exist, there is no longer a provided way to mock endpoints. It's on the user to decide on what client abstraction works best for them, and then use parts provided to construct it themselves, with whatever mocking framework works best for their use case.

It should be *heavily* reviewed and discussed before merging, if it ever is merged, since it changes 99% of the existing API surface.


ObsidianMinor:
I also added an implementation of the List Workers and Upload Worker endpoints for proof of concept for how this new implementation supports multipart endpoints. It doesn't support everything though (just enough for wrangler), so I may take it off and open a separate PR for it later with a much more in-depth implementation.

sssilver:
This is a great body of work; love the motivation and design ideas behind it.

I wonder if instead of it being one giant PR, it could be bite-size small ones, to make the review more thorough. We could create a new branch just for this so the PRs could go against that, and once everything is reviewed and merged, we could bump the major version and cut a release.

Would that be possible? Or would that be too much of a hassle.

ObsidianMinor:
@sssilver I think that's doable. I'll get started on taking this apart for the new branch.

ObsidianMinor:
Oh yeah, I'm going to need someone else with write access to make the branch I'm pushing PRs to.

sssilver:
> Oh yeah, I'm going to need someone with write access to make the branch I'm pushing PRs to.

You now have write access!

ObsidianMinor:
Closing in favor of #153

PR #166, opened 2021-09-28
Wrangler delete endpoints

ocsfrank:
Adds two endpoints for listing a script bindings and for deleting a Durable Object. These two endpoints are used in a Wrangler PR - https://github.com/cloudflare/wrangler/pull/2086.


ocsfrank:
There weren't any last time I checked. I think api.cloudflare.com is a bit outdated as it doesn't include other endpoints either.

nmldiegues:
Structurally looks good.

In terms of endpoint correctness, I'll trust you've checked :) 

@sssilver  has to be the one merging

PR #169, opened 2021-10-18
Ran fmt and clippy

jspspike:
Maybe we can add check to CI?


sssilver:
This looks great! Do you mind squashing the commits before we can merge it?

taylorlee (approved):
lgtm! this is nice to have - I had wondered why things were currently out of sync. squashing would also be nice though.

PR #174, opened 2022-03-08
deps: remove unused slog dependencies

Noah-Kennedy:
We don't use these slog dependencies anywhere. Let's pull them out.

PR #175, opened 2022-03-08
responses: make messages and errors default if not present

Noah-Kennedy:
Both fields are often not returned by certain cloudflare APIs (the Zero Trust API is a good example). Allowing serde to default them if not present allows these APIs to be used with this crate.

### DIFF
--- a/cloudflare/src/framework/response/mod.rs
+++ b/cloudflare/src/framework/response/mod.rs
@@ -9,7 +9,9 @@ use serde_json::value::Value as JsonValue;
 pub struct ApiSuccess<ResultType> {
     pub result: ResultType,
     pub result_info: Option<JsonValue>,
+    #[serde(default)]
     pub messages: JsonValue,
+    #[serde(default)]
     pub errors: Vec<ApiError>,
 }
 


cloudflare/src/framework/response/mod.rs
```diff
@@ -9,7 +9,9 @@ use serde_json::value::Value as JsonValue;
 pub struct ApiSuccess<ResultType> {
     pub result: ResultType,
     pub result_info: Option<JsonValue>,
+    #[serde(default)]
```
adamchalmers, line 12: I think it's more correct to make this `Option<JsonValue>`, rather than defaulting to `JsonValue::Null`. What do you think @sssilver -- one downside is that it would be a breaking change (although a very easy to fix one)

Noah-Kennedy: I agree with @adamchalmers here. The reason I chose to do it this way is that I felt that we could do just a simple patch release with the changes from this and #174 if we kept this nonbreaking.

Noah-Kennedy:
@adamchalmers @sssilver I'd like to get this in. Which approach do we want to go with here?

kennetheversole:
Hello, 

PR #177, opened 2022-03-16
Add endpoint to read values from KV

caass:
Draft for now since I still need to test if this actually works, but `serde_json` supports deserializing raw strings:

```rust
fn main() {
    let hello: String = serde_json::from_str("\"hello\"").unwrap();
    println!("{hello}")
}
```

```sh
   Compiling json-string v0.1.0 (/Users/cass/Projects/work/json-string)
    Finished dev [unoptimized + debuginfo] target(s) in 1.22s
     Running `target/debug/json-string`
hello
```

The [API Docs](https://api.cloudflare.com/#workers-kv-namespace-read-key-value-pair) indicate that KV just returns a raw string, so...it should work? But I'm going to test it out before marking this PR as ready for review.

### DIFF
--- a/cloudflare/src/endpoints/workerskv/mod.rs
+++ b/cloudflare/src/endpoints/workerskv/mod.rs
@@ -9,6 +9,7 @@ pub mod delete_bulk;
 pub mod delete_key;
 pub mod list_namespace_keys;
 pub mod list_namespaces;
+pub mod read;
 pub mod remove_namespace;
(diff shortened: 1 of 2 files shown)


Noah-Kennedy:
@caass you also need to fix some clippy lints

caass:
Ah it was too good to be true anyway, I can't seem to make it work.

Noah-Kennedy:
Ah, sad

PR #180, opened 2022-03-21
Release 0.9.1

nmldiegues:

PR #181, opened 2022-03-21
Release 0.9.2

nmldiegues:
Same as https://github.com/cloudflare/cloudflare-rs/pull/180
0.9.1. did not do anything because I tagged it wrong (it should had been v0.9.1)

### DIFF
--- a/cloudflare/Cargo.toml
+++ b/cloudflare/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "cloudflare"
-version = "0.9.1"
+version = "0.9.2"
 authors = ["Areg Harutyunyan <areg@cloudflare.com>"]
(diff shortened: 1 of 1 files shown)


Noah-Kennedy (changes requested):
We shouldn't bump without a corresponding release. Can you revert the other one and bump back to 0.9.1?

nmldiegues:
> We shouldn't bump without a corresponding release. Can you revert the other one and bump back to 0.9.1?

Why not? Releases are free. This way there's less confusion/entropy as we move forward.

sssilver (approved):

nmldiegues:
Will re-issue a new release with https://github.com/cloudflare/cloudflare-rs/pull/174

PR #184, opened 2022-03-31
Add resolver IP to HttpApiClientConfig

jeff-hiner:
This adds an optional field to HttpApiClientConfig that lets you specify a pre-resolved IP address for the API.

### DIFF
--- a/cloudflare/Cargo.toml
+++ b/cloudflare/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "cloudflare"
-version = "0.10.0-alpha.0"
+version = "0.9.1"
 authors = ["Areg Harutyunyan <areg@cloudflare.com>"]
 edition = "2018"
 description = "Rust library for the Cloudflare v4 API"
--- a/cloudflare/Cargo.toml
+++ b/cloudflare/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "cloudflare"
-version = "0.9.1"
+version = "0.10.0-alpha.0"
 authors = ["Areg Harutyunyan <areg@cloudflare.com>"]
 edition = "2018"
 description = "Rust library for the Cloudflare v4 API"
--- a/cloudflare/src/framework/blocking_api.rs
+++ b/cloudflare/src/framework/blocking_api.rs
@@ -1,5 +1,6 @@
 use reqwest::blocking::RequestBuilder;
 use serde::Serialize;
+use std::net::SocketAddr;
 
 use crate::framework::auth::Credentials;
 use crate::framework::reqwest_adaptors::match_reqwest_method;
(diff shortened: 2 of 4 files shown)


cloudflare/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "cloudflare"
-version = "0.9.1"
```
Noah-Kennedy, line 3: We don't bump this until we do a release PR

Noah-Kennedy (changes requested):

cloudflare/Cargo.toml
Noah-Kennedy: You know what, this procedure can probably change. I think that bumping it to a 0.X+1 alpha in the event of a breaking change is probably a better system overall.

cloudflare/src/framework/blocking_api.rs
```diff
@@ -14,10 +15,19 @@ impl HttpApiClient {
...
-            .default_headers(config.default_headers)
-            .build()?;
+            .default_headers(config.default_headers);
+
+        if let Some(address) = config.resolve_ip {
+            let url = url::Url::from(&environment);
+            builder = builder.resolve(
+                url.host_str()
+                    .expect("Environment url should have a hostname"),
+                SocketAddr::new(address, 443),
```
adamchalmers, line 27: I think the port (443) should be configurable too, what if user want to resolve the given URL with a different port?

sssilver (changes requested):

cloudflare/Cargo.toml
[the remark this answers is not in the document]
Noah-Kennedy: I think we should. They are a very useful tool.

cloudflare/src/framework/blocking_api.rs
jeff-hiner: I added it for clarity, but it's actually ignored by reqwest.
https://docs.rs/reqwest/latest/reqwest/struct.ClientBuilder.html#method.resolve

adamchalmers (approved):

cloudflare/src/framework/blocking_api.rs
adamchalmers: I see, this is _not_ specifying a URL for a DNS resolver, rather, it's letting users skip DNS resolution and instead manually supply a hostname -> IP mapping themselves.

cloudflare/Cargo.toml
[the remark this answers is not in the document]
jeff-hiner: Not a blocker to me either way. I'll revert for now and you can update later, just be aware there's a semver breaking change here.

jeff-hiner:
@sssilver Can you unblock this, please?

sssilver (approved):

PR #187, opened 2022-04-26
framework: support missing messages and/or errors in ApiSuccess

Noah-Kennedy:
A number of services within our infrastructure do not include these fields if they are empty. Although the spec arguably mandates the inclusion of these keys, we should support what happens in the real world rather than what the spec recommends.


jyn514:
Why is it better to change the client than the services? How many services are we talking about? I worry this will hide errors from the services that currently return messages.

adamchalmers (approved):

Noah-Kennedy:
@jyn514 we are talking about a lot of services, some of which I am told are not to be touched. Also some individuals feel strongly that these keys should not be mandated.

PR #194, opened 2022-12-06
Add support delete tunnel with connections

joliveirinha:


cloudflare/src/endpoints/argo_tunnel/delete_tunnel.rs
```diff
@@ -8,6 +8,8 @@ use super::Tunnel;
 pub struct DeleteTunnel<'a> {
     pub account_identifier: &'a str,
     pub tunnel_id: &'a str,
+    // should delete tunnel connections if any exists
+    pub cascade: bool,
```
nmldiegues, line 12: maybe make it optional to not add a "breaking change" to the struct?

nmldiegues (approved):

cloudflare/src/endpoints/argo_tunnel/delete_tunnel.rs
joliveirinha: yes. this is a breaking change. but its a minimal one. 
there is no way around it,afaik

PR #200, opened 2023-01-25
[RUSTSEC-2020-0071] Replace chrono with time 0.3

jeff-hiner:
This is a semver breaking change because the `chrono` types are currently publicly exposed in structs.

PR #201, opened 2023-01-27
Remove time@0.1 transitive dep via chrono

jeff-hiner:
Less disruptive than https://github.com/cloudflare/cloudflare-rs/pull/200 but does fix the issue as long as no other dependency in a project depends on `chrono` with default features.


adamchalmers (approved):

Noah-Kennedy (approved):

keehun:
Can we get a new version for `cloudflare-rs` shipped that includes this change?

PR #213, opened 2023-02-27
listzone may return null id/email for owner field

xgzeng:
I got ListZone response returned from cloudflare with null value for owner.id and owner.email.
The situation causes list zone always fail.

### DIFF
--- a/cloudflare/Cargo.toml
+++ b/cloudflare/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "cloudflare"
-version = "0.12.0"
-authors = ["Noah Kennedy <nkennedy@cloudflare.com>", "Jeff Hiner <jhiner@cloudflare.com>"]
+version = "0.13.0"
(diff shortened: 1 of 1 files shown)


heavycharged:
I'm also meet with this problem. @jeff-hiner can you please help to check and merge this PR?

jeff-hiner:
I'm trying to ask around to see if that's intentional, because I don't think those fields should be blank. Are you seeing this on a specific account?

heavycharged:
Yes, this is on my personal account, as well as on another account that we use on another project. This is a JSON that i dumped before it got parsed:

```json
        ...
        "multiple_railguns_allowed": false
      },
      "owner": {
        "id": null,
        "type": "user",
        "email": null
      },
      "account": {
        "id": "16e2b4933axxxxxxxxxxxxxxxxxxxxxx",
     ...
```

Also, docs states that all of this fields are not required, and therefore can be null:

image

jeff-hiner:
I've asked internally, and this appears to be an API bug. If the `owner` field is present, it should have an `id` and `name`.

heavycharged:
Hmm, interesting. Thanks for investigation. Do you have any ETA when this expected to be fixed?

PR #246, opened 2024-10-02
add repo field to Manifest

DougAnderson444:
It's missing the repo field, which means it's missing from crates.io and docs.rs


DougAnderson444:
Ok! So I'm not the only one....

Issue #251, opened 2025-01-29
Error: error decoding response body: missing field `zone_id` at line 1 column 369


Pablo1107:
> 27th January 2025
> [DNS](https://developers.cloudflare.com/dns/)
> 
> Zone IDs and names on individual DNS records
> 
> Records returned by the API will no longer contain the zone_id and zone_name fields. This change may take up to four weeks to fully roll out. The affected fields were deprecated with an End of Life (EOL) date of November 30, 2024.

[source](https://developers.cloudflare.com/changelog/)

Since 27th January, the API remove some fields and this breaks this library resulting in an error decoding the response body.

Pablo1107:
Resolved by #252

PR #253, opened 2025-02-02
fix(cloudflare): Remove `zone_id` and `zone_name` in DNSRecord

woshiluo:
As https://developers.cloudflare.com/changelog/ said on 27th, Jan, 2025, "Records returned by the API will no longer contain the zone_id and zone_name fields.", so we need to remove it to avoid parse error.

ref #251


newAM (approved):
I tested this out, it fixed #251 for me, thanks!

r-birkner (approved):
Thanks for fixing it!

kennetheversole (approved):
Merging this

PR #258, opened 2025-02-10
fix: removed field auto_added

r-birkner:


kennetheversole:
Can you please provide more context on the issue this was causing?

r-birkner:
We use this crate to create DNS records for the ACME-challenge. With this field, there would be a panic because the crate failed to deserialize the response as the field was missing (it has been deprecated, see [here](https://developers.cloudflare.com/dns/changelog/#2025-02-03)). 

I have noticed that you have overhauled the codebase quite a bit. We are running off of a fork with the applied fixes. I can try switching back to the upstream and see whether it works now.

Riceman2000:
@kennetheversole 
Here is the error message produced when DNS records are listed with version 0.13.0:
```
API request to list existing records failed with: Invalid(reqwest::Error { kind: Decode, source: Error("missing field `auto_added`", line: 1, column: 181) })
```

Thanks @r-birkner for putting this fix together.

GoncaloGarcia:
@r-birkner Could you rebase this please?

ErcinDedeoglu:
?

PR #259, opened 2025-02-12
Add non-JSON response support, multipart body, KV endpoints, etc

### LINKED ISSUE #14: Add `write key` endpoint
https://api.cloudflare.com/#workers-kv-namespace-write-key-value-pair

### LINKED ISSUE #26: HttpApiClient should be able to handle non-JSON request and response bodies

Keiishu:
# Changes
### Add the ability for an endpoint to receive raw data
This was made by:
* adding a flag that defaults to `false`, that represents whether the endpoint responds with raw bytes data or JSON (ApiSuccess);
* adding two types that represents the response type :
	* what the JSON deserialization should expect to get;
	* the actual response type that the user will get.

The presence of those types is necessary if we want users to be able to directly get the actual endpoint response type on their request.

Since Rust doesn't allow default type parameters yet, it is true that we will have to manually set those types for each new endpoint in the future, but I believe it's a fair trade-off for better user experience.
Note that all existing endpoints have already been updated.

### Implement multipart body support
Endpoint bodies can now send `multipart/form-data` content type.

This was implemented by adding a `RequestBody` enum, which is either `Json`, `Raw` or `Multipart`.
The Client's `request` function can then match on `endpoint.body`.

A `MultipartBody` trait had to be created to circumvent the fact that `reqwest`'s `blocking::multipart::Form`/`Part` and `async_impl::multipart::Form`/`Part` aren't inter-compatible, and no `From` or `Into` exists between the two.
It consists of a `parts` function that returns a vector containing the parts of the multipart body. Every endpoint that has the option to send a multipart body has to implement it.

All existing multipart endpoints have already been updated.

### Implement missing endpoints in workerskv
* Get Namespace
* Read Key
* Read Key Metadata
* Write Key

## Refactoring
* Some blocks of code weren't in their associated module or file.
	* Blocking API tests and functions are now in `blocking_api.rs`.
	    * the feature/target_arch check is now in the module definition
	* `ApiResult` is now in `response/mod.rs`.
	* DNS, Zone, Plan, R2 and User endpoints are now in their own module.
	* API Clients are now in their own module.
	* `ApiFailure` tests are now in `api_fail.rs`.
* Change/Rename types according to the official API documentation:
    * `ApiError` is now named `ResponseInfo`, which is the same type as `ApiSuccess.messages`.
* Remove unused feature "endpoint-spec" in `framework/endpoint.rs`.
* Switch from `percent-encoding` to `urlencoding`, which allows us to get rid of `PATH_SEGMENT_ENCODE_SET`.
* Elide lifetimes when possible.

### Docstrings
I updated every docstrings of the `workerskv` module endpoints to reflect the official API documentation.
The other endpoints are still to be done in the future, but I focused on the KV module here.

### Async Client Unit tests
I added unit tests to the async client:
* Successful JSON response handling;
* Successful raw response handling;
* Handling of an endpoint returning an error;
* Test an endpoint that returns nothing;
* Test sending a JSON request body;
* Test sending a raw byte request body;
* Test sending a multipart request;
* Test sending a request with query parameters.

#### Blocking Client
I need help finding a way to use refactor those tests to use them in the blocking client as well, without just copying them over.

#### Mockito Issues
To check test multipart body endpoint, I had to use a regex that apparently wasn't supported by `mockito::Matcher::Regex`, which led to the update of `mockito` to have access to `.match_request()` and check the regex there.

The latest versions of `mockito` uses a server pool. Thus, `mockito::server_url()` doesn't exist anymore.
This isn't an issue for the main crate : I removed `Environment::Mockito`, and users can use `Environment::Custom` to configure the client with `mockito`'s server. But it isn't compatible with current implementation of `cloudflare-examples`.

##### Next steps
I didn't update the `cloudflare-examples` code yet. I can update it in a next PR, but it works as is at the moment, so it isn't a priority.

### Integration Tests
I added an integration test suite for the `workerskv` module.

## Fixes
I added vector length validation on write and delete bulk KV.
I need help finding a way to validate the request size of the write bulk endpoint, as expected and documented in the official API docs.

# Next Steps
While working on this PR, I encountered things I think we should discuss further.
I'm going to create an issue for all of them, but here's a summary if you find it necessary to discuss before squash & merge this PR:

* There are ambiguities in the types, comparing to both the official documentation and what we're actually getting on the endpoint. Should we match the former to align with the docs, or the latter to align with reality?
	* `workerskv::WorkersKvBulkResult`: `unsuccessful_keys` does not seem to be optional. Rather, we just get an empty vector if the list is empty.
	* `workerskv::write_bulk::KeyValuePair` doesn't reflect the official documentation, but neither does this one make sense either. Having everything optional isn't right. What should we do?
	* `zone::Zone`: `plan` doesn't appear in the documentation, neither does it appear in the official TypeScript library. What should we do?
* Argo Tunnels have been renamed to `Cloudflare Tunnels`. `cfd_tunnels` has a lot of the same code as `argo_tunnels`. Should we remove `argo_tunnels` altogether?
* Shouldn't we rename the clients to `AsyncClient` and `BlockingClient`?

# Closing remarks

I mainly focused on working in getting the `workerskv` module complete and working because I needed it for a project, but I'd love to further contribute and see this crate being as complete as the official TypeScript library.

I'm planning on going over each endpoint "group" in the official documentation and implementing them all, module by module. If you have any comments or remarks, don't hesitate to let me know.

@kennetheversole I'm tagging you, after seeing you asked to be tagged in another PR. Thanks in advance for your review and help on this.

Closes #14, closes #26


kennetheversole:
@Keiishu give me a little bit to review this code thoroughly it is chonky. My initial glance is that this is fantastic 

> I'm planning on going over each endpoint "group" in the official documentation and implementing them all, module by module. If you have any comments or remarks, don't hesitate to let me know.

That would be greatly appreciated. I was planning on doing this as well so we should coordinate via the Cloudflare discord.

For future prs please try to keep them contained to a specific API, it make it will for us to get faster review times.

kennetheversole:
@Keiishu Can you do a rebase and pull in the workers ai changes I just merged.

Keiishu:
@kennetheversole For sure, sorry for the chonkiness.
I've joined the Cloudflare Discord, so we can coordinate there. My username is "tinmar_".

I'm rebasing onto `master` and updating the added endpoint according to the changes here.

kennetheversole (approved):
LGTM. Will merge after fixing permissions

Keiishu:
> LGTM. Will merge after fixing permissions

Nice, looking forward to it, thank you for reviewing!
Once it's merged, let's coordinate on Discord for the rest of the endpoints as well as the changes I talked about in my "Next Steps" section!

PR #262, opened 2025-02-19
feat: implement WorkersBinding as an enum 

### LINKED ISSUE #261: Error while deserializing `ListBindings` with bindings other than `kv_namespace`
### Description

The current implementation of `WorkerBinding` assumes that all worker bindings have `namespace_id` as seen here:
https://github.com/cloudflare/cloudflare-rs/blob/f9664e3bd1acbc8cf941be828de5a471613ac829/cloudflare/src/endpoints/workers/mod.rs#L87-L93

Which is only true in case the `WorkerBinding` is a `kv_namespace`.

And it also assume that `name` will be present, which is not the case for `tail_consumer` type.

### Reference

- Binding types: https://developers.cloudflare.com/workers/configuration/multipart-upload-metadata/#bindings

augustoccesar:
Currently the `WorkersBinding` assume that it always have `namespace_id` and also does not take into consideration all the other possible fields of different bindings.
This changes so that the `WorkerBinding` is now an enum with all the possible options listed on https://developers.cloudflare.com/workers/configuration/multipart-upload-metadata/#bindings.

Closes #261

### DIFF
--- a/cloudflare/src/endpoints/workers/mod.rs
+++ b/cloudflare/src/endpoints/workers/mod.rs
@@ -84,13 +84,297 @@ impl ApiResult for WorkersTail {}
 impl ApiResult for Vec<WorkersTail> {}
 
 // Binding for a Workers Script
-#[derive(Deserialize, Serialize, Debug, Clone, PartialEq, Eq)]
-pub struct WorkersBinding {
-    pub name: String,
-    pub r#type: String,
-    pub namespace_id: String,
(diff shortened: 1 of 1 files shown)


kennetheversole:
Can you please rebase?

augustoccesar:
@kennetheversole Done! ✅

kennetheversole (approved):
LGTM

PR #271, opened 2025-03-12
Release v0.14.0

kennetheversole:
Prep Release 0.14.0.

### DIFF
--- a/cloudflare/Cargo.toml
+++ b/cloudflare/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "cloudflare"
-version = "0.13.0"
+version = "0.14.0"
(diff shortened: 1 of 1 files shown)


ErcinDedeoglu:
This blocks the development process, by the way. Do we have an estimate?

Riceman2000:
I also don't want to put undue pressure on you @kennetheversole but what can we do to make this process easier? 
Personally I think issues/PRs over about 6-8 months old could be dropped or marked as stale. This could be automated using bots like: https://github.com/googleapis/repo-automation-bots

Of course I understand if you are limited with what you can do because of the corporate restrictions at Cloudflare. But if you need an additional maintainer I would volunteer for that. 

Thanks for getting this PR up, it seems like you are working this in your spare time so any work that gets done is appreciated.

Noah-Kennedy (approved):

PR #274, opened 2025-05-26
Mark members of AssistantFunction & AssistantTool as pub

ewpratten:
It was priorly impossible to construct these structs due to the private members. I don't see any reason they should have been private in the first place.

FYI @kennetheversole (bugged you internally about this)


kennetheversole (approved):
LGTM

PR #283, opened 2025-11-09
dns: impl PartialEq for DnsContent

newAM:
This is useful for comparing records in dynamic DNS clients.


newAM:
I saw after making this PR that there is a similar open PR #227, but it has merge conflicts so I will leave this open for now.

